PR #2, opened 2022-11-09
Add publish scripts

owenkim:
@ygormsantos 
After following [this](https://medium.com/mobile-app-development-publication/upload-to-mavencentral-made-easy-for-android-library-30d2b83af0c7):

How does this look? I had to deviate from the directions, in that `apply from: "${rootDir}/scripts/publish-root.gradle"` was missing from the directions and I had to change `from components.release` to `from components.prodRelease`. 

I haven't run this yet but was able to confirm that the publish task is available among `gradlew tasks --all`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,10 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
+buildscript {
+    dependencies {
+        classpath 'io.github.gradle-nexus:publish-plugin:1.1.0'
+    }
+}
+
 plugins {
     id 'com.android.application' version '7.3.1' apply false
     id 'com.android.library' version '7.3.1' apply false
@@ -19,3 +25,7 @@ spotless {
         ktlint()
     }
 }
+
+apply plugin: 'io.github.gradle-nexus.publish-plugin'
+
+apply from: "${rootDir}/scripts/publish-root.gradle"
--- a/scripts/publish-module.gradle
+++ b/scripts/publish-module.gradle
@@ -1,0 +1,77 @@
+apply plugin: 'maven-publish'
+apply plugin: 'signing'
+
+task androidSourcesJar(type: Jar) {
+    archiveClassifier.set('sources')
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -1,4 +1,10 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
+buildscript {
+    dependencies {
+        classpath 'io.github.gradle-nexus:publish-plugin:1.1.0'
+    }
+}
+
```
ygormsantos, line 7: nit: this is another way to add plugins to the project, since we have some using the plugins block, we can move it:

```
plugins {
    id 'com.android.application' version '7.3.1' apply false
    id 'com.android.library' version '7.3.1' apply false
    id 'org.jetbrains.kotlin.android' version '1.7.10' apply false
    id 'org.jetbrains.kotlin.plugin.serialization' version '1.7.10'
    id "com.diffplug.spotless" version "6.11.0"
    id 'androidx.navigation.safeargs' version '2.5.3' apply false
    id 'com.google.dagger.hilt.android' version '2.44' apply false
    id 'io.github.gradle-nexus.publish-plugin' version '1.1.0'
}
```

and remove `apply plugin: 'io.github.gradle-nexus.publish-plugin'`

scripts/publish-module.gradle
```diff
@@ -0,0 +1,77 @@
...
+            release(MavenPublication) {
+                // The coordinates of the library, being set from variables that
+                // we'll set up later
+                groupId PUBLISH_GROUP_ID
+                artifactId PUBLISH_ARTIFACT_ID
+                version PUBLISH_VERSION
+
+                // Two artifacts, the `aar` (or `jar`) and the sources
+                if (project.plugins.findPlugin("com.android.library")) {
+                    from components.prodRelease
```
ygormsantos, line 32: Sorry, I should have told you that it's indeed `prodRelease` and not only `release`

PR #9, opened 2022-11-23
Cherry-pick TokenizeCardService

ygormsantos:
# Description
This PR creates TokenizeCardService, and its tests

### DIFF
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -1,6 +1,7 @@
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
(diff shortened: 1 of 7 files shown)


forage-android/build.gradle
```diff
@@ -1,6 +1,7 @@
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
+    id("com.google.devtools.ksp").version("1.7.21-1.0.8")
```
ygormsantos, line 4: KSP is being used only by Moshi to generate the adapter classes on our tests

PR #25, opened 2023-02-22
Add CI workflow

dleis612:
# Description
Created a CI workflow that builds the SDK, runs all unit tests, and runs our automated test suite in browserstack on the sample app.

## Tests Added / Updated?
Yes, tests were added in the form of QA tests.


.github/workflows/CI.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: Run unit tests for the sandboxDebug build
+        run: ./gradlew testSandboxDebugUnitTest --stacktrace
+
+      - name: Build all build variants
+        run: ./gradlew assemble --stacktrace
+
+      - name: Upload App to BrowserStack
+        id: uploadApp
+        uses: fjogeleit/http-request-action@v1
```
ForageRob, line 38: Can you just use cURL instead of this guy's open source action? Definitely not the kind of dependency we want to rely on for security reasons

.github/workflows/CI.yaml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          url: https://api-cloud.browserstack.com/app-automate/upload
+          method: POST
+          username: ${{ secrets.BROWSERSTACK_USERNAME }}
+          password: ${{ secrets.BROWSERSTACK_ACCESS_KEY }}
+          data: '{ "custom_id": "AndroidApp" }'
+          files: '{ "file": "${{ github.workspace }}/sample-app/build/outputs/apk/staging/debug/sample-app-staging-debug.apk" }'
+          timeout: 30000
+
+      - name: Print Output of BrowserStack
```
ForageRob, line 48: Not needed if cURL is logging to console

ForageRob (changes requested):
LGTM, just not comfortable using an open source HTTP request action.

ForageRob (approved):

PR #31, opened 2023-04-20
Implement LaunchDarkly in Android SDK

dleis612:
# Description
Implement the LaunchDarkly SDK and prepare for the BT SDK.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
@@ -1,5 +1,6 @@
 package com.joinforage.forage.android
 
+import android.app.Application
 import android.content.Context
 import com.joinforage.forage.android.collect.VGSPinCollector
 import com.joinforage.forage.android.core.Logger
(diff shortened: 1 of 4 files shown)


forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
```diff
@@ -106,6 +119,10 @@ class ForagePINEditText @JvmOverloads constructor(
         addView(getLogoImageViewLayout(context))
     }
 
+    internal fun getVaultType(): String {
+        return vaultType
+    }
```
djoksimo, line 124: is this going to be used by the consumer of the SDK? I think it's safe to remove this.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -19,21 +20,70 @@ import com.joinforage.forage.android.network.data.CheckBalanceRepository
...
+import com.launchdarkly.sdk.android.integrations.TestData
 import java.util.UUID
 
+internal object VaultConstants {
+    const val VGS_VAULT_TYPE = "vgs_vault_type"
+    const val BT_VAULT_TYPE = "bt_vault_type"
+}
+
+internal object LDConstants {
+    const val VAULT_TYPE_FLAG = "vault-primary-traffic-percentage"
```
djoksimo, line 35: can we refactor `VAULT_TYPE_FLAG` to `VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG` or something alike?

we can consider creating a separate wrapper class for each flag to contain the default value and any mapping / fallback / retrieval logic

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -19,21 +20,70 @@ import com.joinforage.forage.android.network.data.CheckBalanceRepository
...
+                .build()
+        }
+        val context = LDContext.create(LDConstants.USER)
+        val client = LDClient.init(app, ldConfig, context, 0)
+        val vaultPercent = client.doubleVariation(LDConstants.VAULT_TYPE_FLAG, 0.0)
+        val randomNum = Math.random() * 100
+        vaultType = VaultConstants.BT_VAULT_TYPE
+        if (randomNum < vaultPercent) {
+            vaultType = VaultConstants.VGS_VAULT_TYPE
+        }
```
djoksimo, line 72: shouldn't vaultType = VGS_VAULT_TYPE if we set the flag value to 0% (e.g. 100% vgs usage)?

we should also consider the case where we can't connect to LD and default to 0 (e.g. should we default to BT or VGS in the case of being unable to reach LD)?

I think we can safely re-use Shardendu's logic on the backend here:

```kotlin
val randomNum = (1..100).random().toDouble()
if (randomNum < vaultPercent) {
  vaultType = VaultConstants.BT_VAULT_TYPE
} else {
  vaultType = VaultConstants.VGS_VAULT_TYPE
}
```

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
djoksimo: This may make it easier to test and mock the values of flags in our tests.

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
dleis612: Thanks for the calllout, I flipped the logic here: [9f2e3dc](https://github.com/teamforage/forage-android-sdk/pull/31/commits/9f2e3dcfba1559e369e43ded24af68f69a99772d). The default is return BT if anything goes wrong.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
dleis612: Flag name changed and contexts/flags have been separated [9f2e3dc](https://github.com/teamforage/forage-android-sdk/pull/31/commits/9f2e3dcfba1559e369e43ded24af68f69a99772d). In terms of a wrapper around the flag logic, I think we can punt that down the line for now. There may come a time where that is a smart move, but I don't think adding any new flags to the mobile SDKs is very close on the roadmap.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -19,21 +20,70 @@ import com.joinforage.forage.android.network.data.CheckBalanceRepository
...
+        return vaultType as String
+    }
+
+    // IMPORTANT: This function is only used for unit testing. Do not use it in
+    // production code!
+    internal fun reset() {
+        vaultType = null
+    }
```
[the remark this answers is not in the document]
dleis612: This pattern is much better, thanks for the pointer! I went back and forth on whether I should throw an error or not if someone tries to set the `_vaultType` a second time, but left it as a warning for now.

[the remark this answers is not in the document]
djoksimo: Thank you!

djoksimo (approved):
Nice 🚢 🚀 !

PR #37, opened 2023-06-07
[TW-189] Doc updates

kimberleehowley:
# Description

This PR adds parameter definitions. It also includes some minor edits like adding in headers and changing future tense to present. 

Please let me know of any questions or of things that you'd like me to change. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,35 +2,35 @@
 
 Forage Android SDK
 
-Table of contents
-=================
+# Table of contents
 
 <!--ts-->
-   * [Overview](#overview)
-   * [Installation](#installation)
-   * [UI Components](#ui-components)
-     * [ForagePANEditText](#foragepanedittext)
-     * [ForagePINEditText](#foragepinedittext)
-     * [How to provide styles](/docs/styles.md) 
-   * [Tokenizing an EBT Card](#tokenize-ebt-card-number) 
-   * [Performing a balance check](#performing-a-balance-check) 
-   * [Capturing a payment](#capturing-a-payment) 
-   * [The ForageApiResponse sealed class](#the-forageapiresponse-sealed-class)
-   * [Running the Sample App](#running-the-sample-app)
-   * [Dependencies](#dependencies)
+
+- [Overview](#overview)
+- [Installation](#installation)
+- [UI Components](#ui-components)
+  - [ForagePANEditText](#foragepanedittext)
+  - [ForagePINEditText](#foragepinedittext)
+  - [How to provide styles](/docs/styles.md)
+- [Tokenizing an EBT Card](#tokenizing-an-ebt-card)
+- [Performing a balance check](#performing-a-balance-check)
+- [Capturing a payment](#capturing-a-payment)
+- [The ForageApiResponse sealed class](#the-forageapiresponse-sealed-class)
+- [Running the Sample App](#running-the-sample-app)
+- [Dependencies](#dependencies)
 <!--te-->
 
 ## Overview
 
-This documentation covers the integration of Forage's Android SDK to process EBT payments.
+This documentation explains how to integrate the Forage Android SDK to process EBT payments.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -335,7 +356,14 @@ This is an example of usage inside an ACC ViewModel:
 ```
 
 ## Capturing a payment
-ForageSDK will expose the following function to capture a payment:
+
+### Step 1: Add the `ForagePINEditText` UI component
+
+You need to use the `ForagePINEditText` component to your app to check a card's balance. If you need help, refer to the instructions for [adding the UI component](#step-1-add-the-foragepanedittext-ui-component) in the method above. Just change the component name!
+
+### Step 2: Capture the EBT payment
```
ForageRob, line 364: Can we make this step 3 and add a step 2 which is to call a route on the customer's backend which creates the Payment object. The `ref` of that Payment object is an input to the next step and we also got some feedback from customers that server side interactions are not clear in these docs.

README.md
```diff
@@ -249,39 +255,54 @@ ForageSDK will expose the following function to collect the EBT card number:
     ): ForageApiResponse<String>
 ```
 
+#### Parameter definitions
+
+- `merchantAccount`: A unique seven digit numeric string that [FNS](https://docs.joinforage.app/docs/ebt-online-101#food-and-nutrition-service-fns) issues to authorized EBT merchants.
+- `bearerToken`: A [session token](https://docs.joinforage.app/reference/create-session-token).
```
ForageRob, line 261: Let's emphasize here that session token should be created by a route in the customer's backend. The idea is to piggy back on the existing auth between our customers FE and BE.

README.md
```diff
@@ -249,39 +255,54 @@ ForageSDK will expose the following function to collect the EBT card number:
...
             }
             is ForageApiResponse.Failure -> {
                 _error.value = response.message
             }
         }
-    
+
         _isLoading.value = false
     }
 ```
```
ForageRob, line 295: Can we add an optional step 3 here for "persisting the PaymentMethod ref in your wallet". lots of our customers will have wallet functionality in their apps so that users see saved payment methods on their next login. We should note that here as it is a missing piece of backend functionality.

README.md
```diff
@@ -380,11 +416,14 @@ This is an example of usage inside an ACC ViewModel:
         }
 ```
 
-The `paymentRef` will be used to determine if it's a capture from EBT Cash or EBT SNAP. If you choose to support both SNAP and EBT Cash, you will need to handle two payments in your implementation.
+The `paymentRef` is used to determine if it's a capture from SNAP or EBT Cash.
```
dleis612, line 419: This line can probably be removed or the wording should be adjusted. If we want to keep it, I would change it to something like "The `paymentRef` can be used to determine which type of tender is being captured."

dleis612 (changes requested):

README.md
```diff
@@ -249,39 +255,54 @@ ForageSDK will expose the following function to collect the EBT card number:
...
         _isLoading.value = false
     }
 ```
 
 ## Performing a balance check
-ForageSDK will expose the following function to check the EBT card balance:
+
+### Step 1: Add the `ForagePINEditText` UI component
+
+You need to use the `ForagePINEditText` component to your app to check a card's balance. If you need help, refer to the instructions for [adding the UI component](#step-1-add-the-foragepanedittext-ui-component) in the method above. Just change the component name!
```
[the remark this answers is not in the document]
kimberleehowley: Yikes, sorry, thank you for catching these, @dleis612 ! I'm updating now.

README.md
kimberleehowley: I'll adjust for now, but if it reduces complexity for developers to just remove this line, please let me know, @dleis612 ! Thank you.

dleis612 (approved):

owenkim (approved):

PR #38, opened 2023-06-08
Danny/for 1860 android implement a btpincollector class

dleis612:
# Description
Add BasisTheory to the android SDK as a redundant vaulting provider.


forage-android/src/main/java/com/joinforage/forage/android/ui/PINVaultTextField.kt
```diff
@@ -0,0 +1,279 @@
...
+                        // TODO: Consider exposing Padding as an editable field
+                        setPadding(20, 20, 20, 20)
+                    }
+                } finally {
+                    recycle()
+                }
+            }
+    }
+
+    override fun getVGSEditText(): VGSEditText {
```
dleis612, line 108: Change to private

forage-android/src/main/java/com/joinforage/forage/android/ui/PINVaultTextField.kt
```diff
@@ -0,0 +1,279 @@
...
+        return _internalEditText
+    }
+
+    override fun getTextElement(): TextElement {
```
dleis612, line 112: Change to private

forage-android/src/main/java/com/joinforage/forage/android/ui/PINVaultTextField.kt
```diff
@@ -0,0 +1,279 @@
...
+        throw RuntimeException("Unimplemented for this vault!")
+    }
+
+    override fun getUnderlying(): VGSEditText {
```
dleis612, line 116: Change to private

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -0,0 +1,119 @@
...
+            )
+            body = object {
+                val pin = pinForageEditText.getTextElement()
+                val card_number_token = cardToken
+            }
+            path = balancePath(paymentMethodRef)
+        }
+
+        runBlocking {
+            val response = bt.proxy.post(proxyRequest)
```
dleis612, line 40: What happens if this request response with a 400?

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -39,6 +39,7 @@ internal class CapturePaymentRepository(
         return when (val response = paymentService.getPayment(paymentRef)) {
             is ForageApiResponse.Success -> getTokenFromPaymentMethod(
                 paymentRef = paymentRef,
+                // TODO: Parse the token to get BT or VGS
```
dleis612, line 42: Still need to implement token parsing logic

forage-android/build.gradle
```diff
@@ -97,6 +107,14 @@ dependencies {
     // VGS Collect SDK
     api 'com.verygoodsecurity:vgscollect:1.7.3'
 
+    // Basis Theory SDK
+    implementation ('com.github.basis-theory:basistheory-android:2.1.1') {
+        // Unable to build without excluding this dependency
+        // based on advice from this thread: https://github.com/gradle/gradle/issues/3065#issuecomment-341418873
+        exclude group: 'javax.ws.rs'
+    }
+    implementation "javax.ws.rs:javax.ws.rs-api:2.1@jar"
```
ygormsantos, line 116: `com.github.basis-theory:basistheory-java:0.5.0` is depending on `javax.ws.rs-api`.

The last time I integrated BT I didn't have this issue, but it was on version 1.4. Is this expected from Basis Theory?

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -0,0 +1,119 @@
...
+import com.joinforage.forage.android.BuildConfig
+import com.joinforage.forage.android.network.model.ForageApiError
+import com.joinforage.forage.android.network.model.ForageApiResponse
+import com.joinforage.forage.android.network.model.ForageError
+import com.joinforage.forage.android.ui.ForagePINEditText
+import kotlinx.coroutines.runBlocking
+import org.json.JSONException
+import kotlin.coroutines.suspendCoroutine
+
+internal class BTPinCollector(
```
ygormsantos, line 14: We need to change this guy a little bit.

On VGS implementation, they used a really old approach to send the requests. The `suspendCoroutine` function was necessary to transform the callbacks approach into suspending function.

Also, BT documentation didn't help much here; usually, `runBlocking` should be used only for tests.

One approach we can use here is to get rid of the `suspendCoroutine` function since `bt.proxy.post` is already a suspending function:

```
    override suspend fun collectPinForBalanceCheck(
        paymentMethodRef: String,
        cardToken: String,
        encryptionKey: String
        val bt = buildBt()

        val proxyRequest: ProxyRequest = ProxyRequest().apply {
            headers = mapOf(
                "X-KEY" to encryptionKey,
                "Merchant-Account" to merchantAccount,
                "BT-PROXY-KEY" to PROXY_ID,
                "Content-Type" to "application/json"
            )
            body = object {
                val pin = pinForageEditText.getTextElement()
                val card_number_token = cardToken
            }
            path = balancePath(paymentMethodRef)
        }

        pinForageEditText.getTextElement().setText("")

        return runCatching {
            bt.proxy.post(proxyRequest)
        }.fold(
            onSuccess = {
                val forageApiError = ForageApiError.ForageApiErrorMapper.from(it.toString())
                val error = forageApiError.errors[0]

                // our logic to check errors in case the request succeeded
                ForageApiResponse.Failure(listOf(ForageError(400, error.code, error.message)))

            },
            onFailure = {
                // our logic to check errors in case the request failed
                ForageApiResponse.Failure(emptyList())
            }
        )
    }
```

ygormsantos: we can follow this on `collectPinForBalanceCheck` & `collectPinForCapturePayment`

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
```diff
@@ -2,158 +2,88 @@ package com.joinforage.forage.android.ui
...
         orientation = VERTICAL
 
-        vaultType = LDManager.getVaultProvider(context.applicationContext as Application)
-        if (vaultType == VaultConstants.BT_VAULT_TYPE) {
-            // TODO: Do BT stuff!
-        } else if (vaultType == VaultConstants.VGS_VAULT_TYPE) {
-            // TODO: Do VGS stuff!
+        var vaultType = LDManager.getVaultProvider(context.applicationContext as Application)
+        vault = if (vaultType == VaultConstants.BT_VAULT_TYPE) {
+            BTVaultWrapper(context, attrs, defStyleAttr)
```
ygormsantos, line 40: This could be a problem:

I'll need to read the LDClient documentation first, but it takes time to initialize the client and fetch the feature flag. Here we are drawing the UI element; this might freeze the app / generate ANRs.

Probably we might need to create some init method inside `ForageSDK` to fetch this before we reach this part

forage-android/build.gradle
dleis612: Unsure, but I was unable to import any versions of the BT SDK without excluding this dependency. I followed the advice found in the thread commented to fix this and whatever code path we are using through BT doesn't seem to rely on this package.

BT had to make multiple updates and new releases to be compatible with our app in the first place.

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
dleis612: Thank you for the clarification! `suspendCoroutine` was removed here: [7db5c7e](https://github.com/teamforage/forage-android-sdk/pull/38/commits/7db5c7e1321e735be25afc8b46fab634ed71d8ce)

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
dleis612: I have tested this locally and the launch darkly initialization code is executed prior to drawing the component. I haven't noticed any performance issues or app freezing (however it's very possible that this could show up once the code is used on an actual Android device and not an emulator). From a first glance this strategy seems to be working, but if you have more insight please share!

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -0,0 +1,100 @@
...
+                "Content-Type" to "application/json"
+            )
+            body = object {
+                val pin = pinForageEditText.getTextElement()
+                val card_number_token = cardToken
+            }
+            path = balancePath(paymentMethodRef)
+        }
+
+        val response = bt.proxy.post(proxyRequest)
```
ygormsantos, line 37: This part can also throw exceptions; all network errors will probably be thrown here.

That is the main purpose of running it inside the `runCatching`. You can also replace `runCatching` for a regular try-catch too

dleis612: Thank you for the insight! I was able to force BT to throw an error and test that my changes handled that response gracefully in this commit: [ccef382](https://github.com/teamforage/forage-android-sdk/pull/38/commits/ccef382e81787c87e7988bb9860170ccc064e31b).

ygormsantos: anytime =]

djoksimo (approved):

PR #40, opened 2023-06-22
Danny/fx 252 implement token parser and expect 2 encryption aliases

dleis612:
# Description
Dynamically select the correct `encryption_alias` based on the vault provider and include token parser functions that can accept the current token implementation and the new one. The token parser functions are important because in order to release BT, we need to first release the backend changes and all clients should be able to accept the old and new implementations.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
@@ -4,10 +4,10 @@ import com.basistheory.android.service.BasisTheoryElements
 import com.basistheory.android.service.ProxyRequest
 import com.joinforage.forage.android.BuildConfig
 import com.joinforage.forage.android.model.EncryptionKey
+import com.joinforage.forage.android.model.PaymentMethod
 import com.joinforage.forage.android.network.model.ForageApiError
 import com.joinforage.forage.android.network.model.ForageApiResponse
 import com.joinforage.forage.android.network.model.ForageError
-import com.joinforage.forage.android.model.PaymentMethod
 import com.joinforage.forage.android.ui.ForagePINEditText
 import org.json.JSONException
 import java.util.*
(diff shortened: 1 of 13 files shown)


forage-android/src/main/java/com/joinforage/forage/android/model/EncryptionKey.kt
```diff
@@ -3,14 +3,19 @@ package com.joinforage.forage.android.model
 import org.json.JSONObject
 
 internal data class EncryptionKey(
```
djoksimo, line 5: nit: we should consider renaming `EncryptionKey` -> `EncryptionKeys`^

we refer to the var of this type as `encryptionKeys` across this PR

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -25,6 +28,7 @@ internal class BTPinCollector(
                 "X-KEY" to encryptionKey,
                 "Merchant-Account" to merchantAccount,
                 "BT-PROXY-KEY" to PROXY_ID,
+                "IDEMPOTENCY-KEY" to UUID.randomUUID().toString(),
```
djoksimo, line 31: is this related to the encryption key/alias/token changes or are we adding this to support idempotent balance checks in the android SDK?

djoksimo (approved):
Nice stuff!

I've a left a few non-blocking comments

From my understanding, we'll eventually want to add the ability to fallback to a different vault provider when one of the tokens is missing.

E.g. launch darkly says to use BT, but BT token is missing – so use VGS instead.

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
dleis612: Turns out this was required the whole time 👀

PR #43, opened 2023-07-21
[FX-268] Add optional reusable param to tokenizeCard

miketikh:
As described in [this](https://linear.app/joinforage/issue/FX-268/android-add-the-reusable-flag-to-post-payment-method-endpoint) ticket, adding optional reusable param to tokenizeCard calls. Reusable will be set to true by default, maintaining the current behavior.

## Tests Added / Updated
Added test to validate successful response with or without param

NOTE: Calls to tokenize card should now be returning `reusable` as well (see [here](https://linear.app/joinforage/issue/COR-341/include-reusable-in-response-from-createpaymentmethod-view)), so we should update `PaymentMethod` as well as all the fixtures to include that param, but not sure if that falls within the scope of this diff:
```
val response =
            PaymentMethod.ModelMapper.from((paymentMethodResponse as ForageApiResponse.Success).data)
        assertThat(response).isEqualTo(
            PaymentMethod(
                ref = "1f148fe399",
                type = "ebt",
                balance = null,
                card = Card(
                    last4 = "7845",
                    token = "tok_sandbox_sYiPe9Q249qQ5wQyUPP5f7"
                ),
                customerId = "test-android-customer-id",
               // SHOULD HAVE reusable
            )
        )
```

## Screenshots

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
@@ -27,7 +27,8 @@ object ForageSDK : ForageSDKApi {
     override suspend fun tokenizeEBTCard(
         merchantAccount: String,
         bearerToken: String,
-        customerId: String
+        customerId: String,
+        reusable: Boolean
     ): ForageApiResponse<String> {
         val currentEntry = panEntry
 
@@ -41,7 +42,8 @@ object ForageSDK : ForageSDKApi {
                 httpUrl = ForageConstants.provideHttpUrl()
             ).tokenizeCard(
                 cardNumber = currentEntry.getPanNumber(),
-                customerId = customerId
+                customerId = customerId,
+                reusable = reusable
             )
             else -> ForageApiResponse.Failure(listOf(ForageError(400, "invalid_input_data", "Invalid PAN entry")))
         }
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKApi.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKApi.kt
@@ -11,7 +11,8 @@ internal interface ForageSDKApi {
     suspend fun tokenizeEBTCard(
         merchantAccount: String,
         bearerToken: String,
-        customerId: String
+        customerId: String,
+        reusable: Boolean = true
     ): ForageApiResponse<String>
 
     suspend fun checkBalance(
--- a/README.md
+++ b/README.md
@@ -251,7 +251,8 @@ The ForageSDK exposes the following function to collect the EBT card number:
     suspend fun tokenizeEBTCard(
         merchantAccount: String,
         bearerToken: String,
-        customerId: String
+        customerId: String,
+        reusable: Boolean = true
     ): ForageApiResponse<String>
 ```
 
@@ -260,6 +261,7 @@ The ForageSDK exposes the following function to collect the EBT card number:
 - `merchantAccount`: A unique seven digit numeric string that [FNS](https://docs.joinforage.app/docs/ebt-online-101#food-and-nutrition-service-fns) issues to authorized EBT merchants.
 - `bearerToken`: A [session token](https://docs.joinforage.app/reference/create-session-token) that authenticates front-end requests to Forage. To create one, send a server-side request from your backend to the `/session_token/` endpoint.
 - `customerId`: A unique ID for the end customer making the payment. If you use your internal customer ID, then we recommend that you hash the value before sending it on the payload.
+- `reusable`: An optional boolean value indicating whether the same card can be used to make multiple payments, set to true by default.
 
 #### Example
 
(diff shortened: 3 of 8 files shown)


forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -27,7 +27,8 @@ object ForageSDK : ForageSDKApi {
     override suspend fun tokenizeEBTCard(
         merchantAccount: String,
         bearerToken: String,
-        customerId: String
+        customerId: String,
+        reusable: Boolean
```
devinmorgan, line 31: nit: I know that the compiler will enforce this but it could be nice to make explicit that `reusable` is optional with a default value of true here
```suggestion
        reusable: Boolean = true
```

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
devinmorgan, line 1: @dleis612 are we still able to make slight but breaking modifications to some of the method signatures in `ForageSDK`. It would be great if we could do something similar to [Mike's `TokenizeEbtCardNumberOptions` in Forage JS](https://github.com/teamforage/forage-js-sdk/pull/99/files#diff-f546abcaeed153450081bf3c1db991cdad422da80a9a32b5440cfe4bc13f183aR158-R164). In the ideal, both `customerId` and `resuable` args would be part of an optional config argument. This way, all of the optional params  can be grouped in a config param.

devinmorgan (commented):
Mike, this PR looks good. I have some outstanding questions for Danny that I"m hoping he can clarify first before approving. Nice work!

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
djoksimo: flagging here that we're planning on making `customerId` optional for both android and ios: https://linear.app/joinforage/issue/FX-314/make-customerid-optional-in-ios-sdk

devinmorgan: OK, so it sounds like my note about optional whether `customerId` should be optional is already a known concern. Good to know, and thanks for sharing!

forage-android/src/main/java/com/joinforage/forage/android/ForageSDKApi.kt
```diff
@@ -11,7 +11,8 @@ internal interface ForageSDKApi {
     suspend fun tokenizeEBTCard(
         merchantAccount: String,
         bearerToken: String,
-        customerId: String
+        customerId: String,
```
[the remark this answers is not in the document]
devinmorgan: @djoksimo clarified that making `customerId` optional is already a known concern for Android and iOS. I'll make this comment as resolved then since that's out of scope for this PR.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
dleis612: This won't be a breaking change due to being an optional field, however I like where your head is at. Maybe included in the `customerId` change we migrate to an optional config argument. We would then require external devs to be deterministic with the optional params they are using and wouldn't have to continue to adjust the method signature.

dleis612:
Could you update the fixture as well: https://github.com/teamforage/forage-android-sdk/tree/main/forage-android/src/test/resources/fixtures/payment/methods

This may require updating a unit test or two, I forget how we validate the return structure of the tokenize endpoint.

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
devinmorgan: Just created [a ticket](https://linear.app/joinforage/issue/FX-363/android-move-reusable-and-customerid-args-into-optionsconfig) in Linear to capture this.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
miketikh: Hi Devin, I tried to do add this, but it produces the error `an overriding function is not allowed to specify default values for its parameters`. 

For this reason, I figured setting it to true by default in the API it implements would be passed on in this function

devinmorgan (approved):
Everything that I had in mind was addressed - I see that Danny had some feedback as well but approving on my end 🚀

miketikh:
> Could you update the fixture as well: https://github.com/teamforage/forage-android-sdk/tree/main/forage-android/src/test/resources/fixtures/payment/methods
> 
> This may require updating a unit test or two, I forget how we validate the return structure of the tokenize endpoint.
> 
> You will also need to update the README: https://github.com/teamforage/forage-android-sdk/blob/main/README.md#step-2-tokenize-the-ebt-card-number since this is the main source of documentation we have for the Android SDK. I believe there is also mention of the method signature in the docs page: https://docs.joinforage.app/docs/android

The reason I was a bit hesitant before was that updating the tests also requires updating the PaymentMethod class, which should now also contain a reusable field, but didn't want to accidentally affect any other fns using this. I updated the class to contain an optional reusable field and all the tests pass, so hopefully no issues from that.

Also updated the fixtures and readme

dleis612 (approved):

PR #45, opened 2023-07-27
Update the state IIN list

dleis612:
# Description
Gaby found two discrepancies in length that are updated here. These have both been confirmed by checking balance checked cards in our system.


miketikh:
Is this a recent change? Both those states have pan length listed as 19 in our [input-iframes-codes](https://github.com/teamforage/forage-input-iframes/blob/31d0c10518ef0bc6ce9a353c7e7ce97d0a0b5de6/src/store/user-input/state_iin_codes.ts#L4) and on [this site](https://www.nacha.org/sites/default/files/2019-05/State-IINs-04-10-19.pdf)

miketikh (approved):

djoksimo (approved):

dleis612:
> Edit: Nevermind, see you edited the iframes as well, and that website is wrong.
> 
> Is this a recent change? Both those states have pan length listed as 19 in our [input-iframes-codes](https://github.com/teamforage/forage-input-iframes/blob/31d0c10518ef0bc6ce9a353c7e7ce97d0a0b5de6/src/store/user-input/state_iin_codes.ts#L4) and on [this site](https://www.nacha.org/sites/default/files/2019-05/State-IINs-04-10-19.pdf)

Yeah sadly the government document we always referenced as the source of truth (that NACHA website) seems to be incorrect :(

PR #50, opened 2023-08-02
Expose user-defined onFocus and onBlur listeners

devinmorgan:
# Description
The SDK now provides several ways to interact with the focus state of a `ForagePINEditText`:

* **To programmatically request focus**, simply call the `.requestFocus()` method on your `ForagePINEditText` instance. For example:

```kotlin
  yourForagePINEditTextInstance.requestFocus()
```
This will make the `ForagePINEditText` request focus.

* **To query the current focus state**, use the `.getElementState().isFocused` chain of calls on your `ForagePINEditText` instance. For example:
```kotlin
val isFocused = yourForagePINEditTextInstance.getElementState().isFocused
```
isFocused will be `true` if the `ForagePINEditText` has focus, and `false` otherwise.

* **To subscribe to focus events**, use the `.setOnFocusEventListener()` method on your `ForagePINEditText` instance. You can pass a lambda function as an argument to this method, and this function will be called when a focus event occurs. For example:

```kotlin
yourForagePINEditTextInstance.setOnFocusEventListener {
// Handle the focus event here
}
```
* **To subscribe to blur events**, use the `.setOnBlurEventListener()` method on your `ForagePINEditText` instance. Like with focus events, you can pass a lambda function to this method, and it will be called when a blur event occurs. For example:

```kotlin
yourForagePINEditTextInstance.setOnBlurEventListener {
// Handle the blur event here
}
```

## Notes
A lot happened in this commit name:
* introduces the `getElementState` method on ForageUI as a mechanism to
  query the current state of a ForageElement
* introduces the `setOnFocusEventListener` and `setOnBlurEventListener`
  methods as a mechanism to register a listener for when a focus and
  blur event occurs

The rationale is for getElementState is that apparently hasFocus() was
no reliable. During manual testing, it was not accurately reflecting
when the input was focused or not. Likely that was because we were not
adhering to the interface [that Basis Theory exposes](https://developers.basistheory.com/docs/sdks/mobile/android/events#focusevent).
Also, `isFocused` already means something in Android so it would be
against convention to hijack that name for Forage's purposes. Instead, I
took a [hint from what VGS did](https://www.verygoodsecurity.com/docs/vgs-collect/android-sdk/field-state-tracking/#:~:text=The%20getAllStates%20method%20returns%20all%20states%20of%20views%2C) which is to expose a method to query state
directly, this way avoiding hijacking of the already used `.isFocused`
field.

Other things of note in the code comments is that we needed to use
mutable pointers to the event subscribers that people will pass in order
to smooth out the differences between how BT and VGS implement their
respective event listeners

## Tests Added / Updated?
I manually tested all of these changes for both VGS and Basis Theory and they appear to work fine

## Screenshots


devinmorgan:
* main
  * **PR #50** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/50" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/50?utm_source=stack-comment).

djoksimo (approved):

PR #51, opened 2023-08-03
[POP-69] Enforce 80% code coverage for PRs

nestorga:
# Description
This PR adds a new GitHub Action workflow to implement test coverage checks. I considered adding this to an existing workflow, but thought it cleaner to keep it isolated. Also, it appears someone tried this in the past as [9 months ago](https://github.com/teamforage/forage-android-sdk/actions/workflows/Coverage.yaml) there was a `Coverage.yaml`.

Coverage is enforced at 80% via a two-step process that can be extended to other repos:

- First, an XML code coverage report is generated by kover. Any coverage reports can be used so long as they are in Cobertura, Clover, JaCoCo or LCov format.
- Next, we use the [diff_cover](https://pypi.org/project/diff-cover/) package to parse the XML output and compare the current PR to `origin/main`. If coverage for added lines is below 80%, the command will return nonzero exit status and fail the step.

Once this is done, we can enforce the `coverage` job as a `main` branch status check. 

## Tests Added / Updated?
- This is GitHub Actions, so we have to test it live on this PR. If we merge, it will apply to future PRs. 
- I verified locally that the coverage checking functionality works as expected.


.github/workflows/Coverage.yaml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'zulu'
+          java-version: 11
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+
+      - name: Setup diff_cover
+        run: pip install diff_cover==7.7.0
```
nestorga, line 28: Went down a rabbit hole of trying to pin to a hash here and decided to timebox it (need to run to an interview).

I'm not sure how much SHA pinning here matters if these are public repos.

nestorga: Should I even be pinning a version here?  Could do `>7,<8`.

.github/workflows/Coverage.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+      - name: Generate XML coverage report (JaCoCo)
+        run: ./gradlew forage-android:koverXmlReport
+
+      - name: Verify coverage
+        run: |
+          diff-cover \
+            forage-android/build/reports/kover/xml/report.xml \
+            --src-roots "forage-android/src/main/java" \
+            --compare-branch ${{ github.event.pull_request.base.ref }} \
```
nestorga, line 41: This is nice for stacked PRs.

devinmorgan:
Hey Nicole! This PR is going to be super useful! There's some urgent changes that Forage X needs to ship to GoPuff so it would be great if we started enforcing test coverage (i.e. merge this PR in) either next week or the week after. Really excited for when this PR lands!

nestorga:
Sure, will hold off merging this until you give the go ahead!

nestorga:
@devinmorgan We can merge this but not actually enforce it as a status check until you give the go-ahead!

devinmorgan:
OK if it's an option to merge in without enforcing yet then sounds good!

My worry was that there's a lot of logic in the view layer of these apps with no test coverage, and making tweaks to that logic in the view-layer in earnest for GoPuff I fear would trip the zero-coverage rule. We're adding new logic as standalone services that are much easier to test so the enforcement would be welcomed there. Sharing that context just to be fully transparent

.github/workflows/Coverage.yaml
devinmorgan: thanks for spending your time saving us time! 💛

devinmorgan (approved):
Thanks for trying to make the lives of Forage X better!

PR #58, opened 2023-08-10
Fix focus/blur event handler not firing bug

devinmorgan:
# Description
[Gopuff flagged that the focus and blur events were not firing
correctly](https://joinforage.slack.com/archives/C04FQM5F2DA/p1691443021122609) when they set end-icon-mode to `clear_text`. Really
really really bizarre bug. Regretfully, I still don't know with
certainty what the underlying issue was. For now, the fix seems
to be attaching the event listener after the view has been added

## Tests Added / Updated?
- YES or NO. If not, why?
I tried to setup [Espresso tests](https://developer.android.com/training/testing/espresso) for this but it was taking longer than I could afford and wrote that off as out-of-scope given the urgency of fixing this but

## Screenshots
A video of the `focus`, `blur`, and `change` events firing when the `TextInputLayout` has end-icon-clear-text, which is the setup that was causing GoPuff issues

https://github.com/teamforage/forage-android-sdk/assets/12377418/a7c3c12b-cef8-4047-871a-b947fdb72672

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt
@@ -89,9 +89,6 @@ class ForagePANEditText @JvmOverloads constructor(
         disableCopyCardNumber()
 
         textInputEditText.addTextChangedListener(this)
-        textInputEditText.setOnFocusChangeListener { _, hasFocus ->
-            manager.changeFocus(hasFocus)
-        }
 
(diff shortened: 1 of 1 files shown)


devinmorgan:
* main
  * **PR #58** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/58" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/58?utm_source=stack-comment).

dleis612 (approved):

PR #62, opened 2023-08-14
[BREAKING] `tokenizeEBTCard` on `ForageSDK` accepts `ForagePANEditText` as arg

devinmorgan:
I accidentally closed #61. This is a duplicate if you are experiencing dejavu

# Description
Previously `tokenizeEBTCard`'s implementation relied on
the sharing of `panEntry` within the `ForageSDK` singleton
object. Sharing state is generally considered bad practice
because it does not scale to multiple parallel consumptions
well.

This commit decouples (and deletes) the shared state. Now,
`tokenizeEBTCard` accepts a ForagePANEditText as an argument
and has no reliance on shared state. The end result is:
* easier to test the `shouldTokenize` logic since it can be
  extract to a service and not live in the view layer near the
  shared state
* no shared state means multiple instances of ForagePANEditText
  can co-exist without issue

# Also
- [x] update the Sample App to pass the `ForagePANEditText` element. Will do that in a subsequent commit
- [x] update the README.md to describe new usage of `ForageSDK.tokenizeEBTCard`
- [ ] Add video recording showcasing flow still works as expected

## Tests Added / Updated?
- YES - this PR adds service layer tests for the guard logic of whether to tokenize an EBT card
- NO - we still don't have espresso enabled so no end-2-end tests to validate new view logic. Instead I'll provide screenshots

## Screenshots

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ The ForageSDK exposes the following function to collect the EBT card number:
 ```kotlin
     suspend fun tokenizeEBTCard(
         merchantAccount: String,
+        panForageEditText: ForagePANEditText,
         bearerToken: String,
         customerId: String,
         reusable: Boolean = true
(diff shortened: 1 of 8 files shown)


devinmorgan:
Closing this PR to create a new one associated with the branch `devin/fx-423-android-remove-storeentry-and-panentry-related-code` so that the Linear bot can track this

PR #63, opened 2023-08-14
[FX-443] Expose version in build config and add to logs

dleis612:
# Description
Add the Forage SDK version to the build config and append it to all logs.

Screenshot 2023-08-14 at 7 55 15 PM

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
@@ -54,7 +54,8 @@ internal interface Log {
                     .setLoggerName(LOGGER_NAME)
                     .build()
 
-                logger?.addAttribute("version", BuildConfig.VERSION)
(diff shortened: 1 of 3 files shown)


forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
```diff
@@ -53,6 +53,8 @@ internal interface Log {
                     .setBundleWithTraceEnabled(true)
                     .setLoggerName(LOGGER_NAME)
                     .build()
+
+                logger?.addAttribute("version", BuildConfig.VERSION)
```
dleis612, line 57: @devinmorgan @djoksimo Any thoughts on naming this something like "android_sdk_version" or something? Or are we cool with "version" being the naming convention.

djoksimo: I'm cool with "version"

djoksimo (changes requested):

djoksimo (approved):

PR #64, opened 2023-08-15
Android Sentry initialization

anusha-bharat:
# Description
Sentry initialisation for android to enable error reporting
Project linked  [here](https://team-forage.sentry.io/projects/android-sdk/?project=4505704106688512
)
## Also
- Add any other relevant change made on this PR.

## Notes
- Describe your changes

## Tests Added / Updated?
Tried adding sentry.capture exception and forcing an exception in the PIN and PAN constructors

## Screenshots

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
@@ -63,7 +63,14 @@ internal interface Log {
                     options.tracesSampleRate = 1.0
                     options.isEnableUserInteractionTracing = true
                     options.environment = BuildConfig.FLAVOR
-        }
+                }
+                logger = Logger.Builder()
+                    .setNetworkInfoEnabled(true)
+                    .setLogcatLogsEnabled(false)
+                    .setDatadogLogsEnabled(true)
+                    .setBundleWithTraceEnabled(true)
+                    .setLoggerName(LOGGER_NAME)
+                    .build()
             }
 
             override fun d(msg: String, attributes: Map<String, Any?>) {
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
@@ -8,10 +8,11 @@ import com.datadog.android.log.Logger
 import com.datadog.android.privacy.TrackingConsent
 import com.joinforage.forage.android.BuildConfig
 import io.sentry.android.core.SentryAndroid
+import io.sentry.Sentry
 
 internal interface Log {
     fun initializeDD(context: Context)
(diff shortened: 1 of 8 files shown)


sample-app/build.gradle
```diff
@@ -89,7 +89,7 @@ dependencies {
     implementation 'com.squareup.okhttp3:logging-interceptor:4.10.0'
     implementation 'com.github.skydoves:sandwich:1.3.2'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.6.4'
-
+    implementation 'io.sentry:sentry-android:6.25.2'
```
dleis612, line 92: You don't need to add sentry to the sample app! It will be injected by the Forage SDK when that is installed :)

anusha-bharat: f5a9955925fbf82977bf3c44016df03e4dd2ce9c

djoksimo (commented):
Should we use the [OkHTTP integration](https://docs.sentry.io/platforms/android/configuration/integrations/okhttp/) @dleis612  ?

> so that it will automatically add a breadcrumb and start a span out of the active span bound to the scope for each HTTP Request.

djoksimo (changes requested):
Can we test whether uncaught exceptions/crashes get reported to Sentry? 

```kotlin
throw RuntimeException("This is a crash")
```

djoksimo (changes requested):
Thank you for adding Sentry, @anusha-bharat 

Can we test whether `Sentry.captureMessage("test android-sdk error")` works?

forage-android/src/main/java/com/joinforage/forage/android/core/Log.kt
```diff
@@ -55,6 +58,17 @@ internal interface Log {
                     .build()
             }
 
+            override fun initializeSentry(context: Context){
+                val flavor = BuildConfig.FLAVOR
+                if (flavor == "sandbox" || flavor == "prod"){
```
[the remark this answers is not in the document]
anusha-bharat: Removed the check since we do not want to restrict environments to sandbox and prod. Unit testing would be under the ` getSilentInstance()` instance which does not initialise sentry

anusha-bharat:
> Thank you for adding Sentry, @anusha-bharat
> 
> Can we test whether `Sentry.captureMessage("test android-sdk error")` works?

Tested both locally and on android emulator

anusha-bharat:
> Can we test whether uncaught exceptions/crashes get reported to Sentry?
> 
> ```kotlin
> throw RuntimeException("This is a crash")
> ```

As discussed on slack huddle crashes and undefined exceptions alert on Sentry only when the app is reopened. That is expected behavior based on sentry docs

djoksimo (approved):

PR #67, opened 2023-08-16
[FX-404] Support dynamic PAN masking based on StateIIN

devinmorgan:
# Description
1. Introduces the `FormatPanTextWatcher` 
2. Updates the `ForagePANEditText` to use the `ForagePanTextWatcher` to format its input

## Also
- [x] ☠️  I need to update `ForagePANEditText.afterTextChanged` only to use the raw digits from the input. Since all cardNumbers are formatted, they include whitespace so they will always be marked as `Invalid` which will break things.

## Notes
- See [linear ticket](https://linear.app/joinforage/issue/FX-404/android-support-dynamically-updating-the-mask-of-ebt-card-numbers)
- ⚠️  Currently, formatting the PAN text makes the `ForagePANEditText` no longer a Numbers-only input, so adding formatting results in the regular keyboard popping up instead of the numbers-only keyboard that was showing up before.

## Tests Added / Updated?
- YES - this PR added ~30 unit tests to ensure that formatting works in many different scenarios

## Screenshots
Demo showing:
1. 4{14} white listed cards are Valid
2. 5{14} white listed cards are Valid
3. 9{4} white listed cards are Valid
4. 654321 white listed cards are Valid
5. State IIN Validation 
6. 16 PAN formatting for validate State IIN
7. 18 PAN formatting for validate State IIN
8. 19 PAN formatting for validate State IIN

https://github.com/teamforage/forage-android-sdk/assets/12377418/0bb57ee7-7fec-417d-8b3b-24b3a02b3164

### DIFF
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -126,6 +126,7 @@ dependencies {
     implementation "javax.ws.rs:javax.ws.rs-api:2.1@jar"
 
     implementation "com.datadoghq:dd-sdk-android:1.19.2"
+    implementation 'androidx.test:core-ktx:1.5.0'
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'com.squareup.okhttp3:mockwebserver:4.9.3'
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -126,7 +126,7 @@ dependencies {
     implementation "javax.ws.rs:javax.ws.rs-api:2.1@jar"
 
     implementation "com.datadoghq:dd-sdk-android:1.19.2"
-    implementation 'androidx.test:core-ktx:1.5.0'
+    testImplementation 'androidx.test:core-ktx:1.5.0'
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'com.squareup.okhttp3:mockwebserver:4.9.3'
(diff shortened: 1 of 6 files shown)


devinmorgan:
* main
  * **PR #67** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/67" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/67?utm_source=stack-comment).

forage-android/build.gradle
```diff
@@ -121,6 +121,7 @@ dependencies {
     implementation "javax.ws.rs:javax.ws.rs-api:2.1@jar"
 
     implementation "com.datadoghq:dd-sdk-android:1.19.2"
+    testImplementation 'androidx.test:core-ktx:1.5.0'
```
devinmorgan, line 124: Needed to add this in order to use [Robolectric](https://robolectric.org/) to create JVM unit tests for the `FormatPanTextWatcher`.

djoksimo (changes requested):
`isValid` should default to true (even on empty string) until we see the full IIN, but it doesn't seem to be on the latest commit

djoksimo (changes requested):
seems like we're emitting duplicate change events

The later event (of the two events) is accurate. Maybe we can tolerate this and release but if it's an easy fix, we should make sure we only send one validation event on a single key press

devinmorgan:
> `isValid` should default to true (even on empty string) until we see the full IIN, but it doesn't seem to be on the latest commit

Fixed in this in #70. Made a separate PR to keep changes organized

devinmorgan:
> seems like we're emitting duplicate change events
> 
> The later event (of the two events) is accurate. Maybe we can tolerate this and release but if it's an easy fix, we should make sure we only send one validation event on a single key press

Fixed this in #71 !

djoksimo (approved):
LGTM! We should still merge down the other PRs and manually test all of the combined changes (#70 , ...) before releasing the next version of the android-sdk

devinmorgan:
> LGTM! We should still merge down the other PRs and manually test all of the combined changes (#70 , ...) before releasing the next version of the android-sdk

Yeah that's a fair point. I'll have to do that on `main` I already started merging the stack. Was hoping to not have to squash all the PRs into one squash

PR #68, opened 2023-08-17
[BREAKING] `tokenizeEBTCard` on `ForageSDK` accepts `ForagePANEditText` as arg

devinmorgan:
I accidentally closed #61. This is a duplicate if you are experiencing dejavu

# Description
Previously `tokenizeEBTCard`'s implementation relied on
the sharing of `panEntry` within the `ForageSDK` singleton
object. Sharing state is generally considered bad practice
because it does not scale to multiple parallel consumptions
well.

This commit decouples (and deletes) the shared state. Now,
`tokenizeEBTCard` accepts a ForagePANEditText as an argument
and has no reliance on shared state. The end result is:
* easier to test the `shouldTokenize` logic since it can be
  extract to a service and not live in the view layer near the
  shared state
* no shared state means multiple instances of ForagePANEditText
  can co-exist without issue

# Also
- [x] update the Sample App to pass the `ForagePANEditText` element. Will do that in a subsequent commit
- [x] update the README.md to describe new usage of `ForageSDK.tokenizeEBTCard`
- [x] Add video recording showcasing flow still works as expected
- [x] Check to see if its possible to write tests with Robolectric given that we've got it working in #67 since it would be good to add automated tests where possible
      ❌ Robolectric is out of scope for right now

## Tests Added / Updated?
- YES - this PR adds service layer tests for the guard logic of whether to tokenize an EBT card
- NO - we still don't have espresso enabled so no end-2-end tests to validate new view logic. Instead I'll provide screenshots

## Screenshots

https://github.com/teamforage/forage-android-sdk/assets/12377418/fbfd7684-6f51-445a-bb79-39cbfa33681a


README.md
```diff
@@ -268,11 +270,12 @@ The ForageSDK exposes the following function to collect the EBT card number:
 This is an example of usage inside an ACC ViewModel:
 
 ```kotlin
-    fun onSubmit() = viewModelScope.launch {
+    fun onSubmit(panForageEditText: ForagePANEditText) = viewModelScope.launch {
         _isLoading.value = true
 
         val response = ForageSDK.tokenizeEBTCard(
             merchantAccount = merchantAccount,
+            panForageEditText = ForagePANEditText,
```
djoksimo, line 278:
```suggestion
    fun onSubmit() = viewModelScope.launch {
        _isLoading.value = true
        
        val tokenizeForagePanEditText = root?.findViewById<ForagePANEditText>(R.id.tokenizeForagePanEditText)

        val response = ForageSDK.tokenizeEBTCard(
            merchantAccount = merchantAccount,
            panForageEditText = tokenizeForagePanEditText,
```

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -12,24 +12,24 @@ import com.joinforage.forage.android.network.TokenizeCardService
...
  * Singleton responsible for implementing the SDK API
  */
 object ForageSDK : ForageSDKApi {
-    private var panEntry: String = ""
     private val logger = Log.getInstance()
 
+    // TODO: this should be a Config argument that uses the builder pattern
     override suspend fun tokenizeEBTCard(
         merchantAccount: String,
+        panForageEditText: ForagePANEditText,
```
djoksimo, line 28: thoughts on using the `forage` prefix (consistent with all of our SDKs and match var name closer to its type?

```suggestion
        foragePanEditText: ForagePANEditText,
```

djoksimo (changes requested):

README.md
devinmorgan: Maybe it can be done more than one way depending on the developers preference? I updated the README.md here with the pattern that I used in `sample-app/src/main/java/com/joinforage/android/example/ui/complete/flow/tokenize/FlowTokenizeFragment.kt`.

Let me know if there's something I'm missing that makes one way preferable!

djoksimo: I think your approach is fine. It seems like you're passing the type on line 278, instead of the actual param.

```kotlin
            panForageEditText = ForagePANEditText,
````

Adding this line could make it immediately apparent that the "tokenizeForagePanEditText" is a view and not a regular object – given our [other snippets showcase XML](https://github.com/teamforage/forage-android-sdk#foragepanedittext). Your snippet LGTM too, so feel free to ignore this suggestion

```kotlin
        val tokenizeForagePanEditText = root?.findViewById<ForagePANEditText>(R.id.tokenizeForagePanEditText)
```

devinmorgan: this is a good idea addition! Will do

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
```diff
@@ -106,6 +106,10 @@ class PanElementStateManager(state: ElementState, private val validators: Array<
         onChangeEventListener?.invoke(getState())
     }
 
+    fun canTokenizePanElementValue(alwaysAllow: Boolean = false) : Boolean {
```
dleis612, line 109: Saw that this was removed later, but just want to confirm that we definitely don't want to be restricting the submit function on our side. Let the user of the SDK make that decision.

dleis612 (approved):

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
devinmorgan: We're 100% on the same page! This method's only reference is `ForagePANEditText.shouldTokenize`, which has been [unused since PR 72](https://github.com/teamforage/forage-android-sdk/pull/72/files#diff-a78535f5e62364effd45eeb5e9b9abc64306d9e92f769f380195b6c359171306L45). As you pointed out, one of the PRs in this stack, we delete this method since it has no use. Let me know if that clarifies!

README.md
devinmorgan: Hey Danilo! This thread was not lost on me! I opened up a new PR to keep the README.md in sync with the latest changes, and [I believe I incorporated this feedback in the snippets](https://github.com/teamforage/forage-android-sdk/pull/105/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R261). Lmk any thoughts or feedback you on this and thanks again for surfacing this!

PR #69, opened 2023-08-17
[FX-456] Support Whitelisted PAN Cards in Non-Prod environments

devinmorgan:
# Description
1. Reverted the changes done in #65 
2. Adds support for whitelisted EBT PANs for `ForagePANEditText` when environment is not "prod"

## Also
The commits are very detailed so read one by one for more context.

## Tests Added / Updated?
- YES - adds many dozens of unit tests to verify the whitelisting and Strict IIN validation logic

## Screenshots
Video demo of the whitelisted cards and StateIIN cards being accepted as valid

https://github.com/teamforage/forage-android-sdk/assets/12377418/53eed848-b377-4120-aff8-36a67460d662

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
@@ -14,6 +14,7 @@ import com.joinforage.forage.android.network.TokenizeCardService
 import com.joinforage.forage.android.network.data.CapturePaymentRepository
 import com.joinforage.forage.android.network.data.CheckBalanceRepository
 import com.joinforage.forage.android.network.model.ForageApiResponse
+import com.joinforage.forage.android.network.model.ForageError
 import com.joinforage.forage.android.ui.ForagePINEditText
 import java.util.UUID
 
@@ -40,19 +41,35 @@ object ForageSDK : ForageSDKApi {
             )
         )
 
-        return TokenizeCardService(
-            okHttpClient = OkHttpClientBuilder.provideOkHttpClient(
-                bearerToken,
-                merchantAccount,
-                idempotencyKey = UUID.randomUUID().toString()
-            ),
-            httpUrl = ForageConstants.provideHttpUrl(),
-            logger = logger
-        ).tokenizeCard(
-            cardNumber = currentEntry.getPanNumber(),
-            customerId = customerId,
-            reusable = reusable
-        )
+        return when {
+            shouldTokenize(currentEntry) -> TokenizeCardService(
+                okHttpClient = OkHttpClientBuilder.provideOkHttpClient(
+                    bearerToken,
+                    merchantAccount,
+                    idempotencyKey = UUID.randomUUID().toString()
+                ),
+                httpUrl = ForageConstants.provideHttpUrl(),
+                logger = logger
+            ).tokenizeCard(
+                cardNumber = currentEntry.getPanNumber(),
+                customerId = customerId,
+                reusable = reusable
+            )
+            else -> {
+                logger.e(
+                    "PAN entry was invalid",
+                    attributes = mapOf(
+                        "merchant_ref" to merchantAccount,
+                        "customer_id" to customerId
+                    )
+                )
+                ForageApiResponse.Failure(listOf(ForageError(400, "invalid_input_data", "Invalid PAN entry")))
+            }
+        }
+    }
+
+    private fun shouldTokenize(panEntry: PanEntry): Boolean {
+        return panEntry is PanEntry.Valid || BuildConfig.DEBUG
     }
 
     override suspend fun checkBalance(
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
@@ -7,6 +7,7 @@ import com.joinforage.forage.android.core.element.TooLongEbtPanError
 import com.joinforage.forage.android.model.STATE_INN_LENGTH
 import com.joinforage.forage.android.model.StateIIN
 
+const val MIN_CARD_LENGTH = 16
 const val MAX_CARD_LENGTH = 19
 
 private fun missingStateIIN(cardNumber: String): Boolean {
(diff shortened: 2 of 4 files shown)


devinmorgan:
* main
  * **PR #69** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/69" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/69?utm_source=stack-comment).

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
```diff
@@ -56,27 +68,82 @@ class PanElementStateManager(state: ElementState) : ElementStateManager(state) {
...
+    override fun checkForValidationError(cardNumber: String): ElementValidationError? {
+        // this is a developer-only validator and should remain
+        // quiet by not returning a validationError
+        return null
     }
+}
+
+class PaymentCaptureErrorCard : WhitelistedCards("4", 14)
+class BalanceCheckErrorCard : WhitelistedCards("5", 14)
+class NonProdValidEbtCard : WhitelistedCards("9", 4)
```
djoksimo, line 97: we should support [this case](https://joinforage.slack.com/archives/C056QGRF6S3/p1692290385036899) as well

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
devinmorgan: excellent callout! Will add shortly!

devinmorgan: Address in [7a75444bb](https://github.com/teamforage/forage-android-sdk/pull/69/commits/7a75444bb050472b3348eb784e290af37aaf412c)

djoksimo (changes requested):
It seems like 16-digit cards that start with special prefixes (9999, 4444...) are only considered complete (`isComplete = true`) when they are 19-digits long. They should be considered complete when 16 <= PAN length <= 19 for these special cards.

The latest specification can be found here: https://linear.app/joinforage/issue/FX-463/disable-js-validation-for-special-cards-in-non-prod-environments

devinmorgan:
> It seems like 16-digit cards that start with special prefixes (9999, 4444...) are only considered complete (`isComplete = true`) when they are 19-digits long. They should be considered complete when 16 <= PAN length <= 19 for these special cards.
> 
> The latest specification can be found here: https://linear.app/joinforage/issue/FX-463/disable-js-validation-for-special-cards-in-non-prod-environments

OK I have no issue with that requirement. Let me make that now

djoksimo (approved):
LGTM!

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -40,19 +41,35 @@ object ForageSDK : ForageSDKApi {
...
+                        "merchant_ref" to merchantAccount,
+                        "customer_id" to customerId
+                    )
+                )
+                ForageApiResponse.Failure(listOf(ForageError(400, "invalid_input_data", "Invalid PAN entry")))
+            }
+        }
+    }
+
+    private fun shouldTokenize(panEntry: PanEntry): Boolean {
```
dleis612, line 71: Remove this function and always let someone attempt to tokenize the EBT card.

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
```diff
@@ -74,31 +69,84 @@ class PanElementStateManager(state: ElementState) : ElementStateManager(state) {
...
     companion object {
         fun forEmptyInput(): PanElementStateManager {
-            return PanElementStateManager(INITIAL_ELEMENT_STATE)
+            return PanElementStateManager(
+                INITIAL_ELEMENT_STATE,
+                arrayOf(StrictEbtValidator())
+            )
+        }
+
+        fun DEV_ONLY_forEmptyInput(): PanElementStateManager {
```
dleis612, line 139: I see this being used in tests as well as a check to use this whenever we aren't in `prod`. If this is entirely for tests, leaving it as is makes sense. If this is the function that should be run if we aren't in production, could we change the name? I'm not sure what is best, but `NON_PROD_forEmptyInput`? What do you think?

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
devinmorgan: No issue with changing the name to `NON_PROD_forEmptyInput`! And, apologies, I don't think I fully grasped your whole remark but `NON_PROD_forEmptyInput` is tested in tests, but its main usage is in `ForagePANEditText` [when instantiating the `manager` property](https://github.com/teamforage/forage-android-sdk/pull/69/files/554202fe522df990691dc583e5d970ca5719e8f8#diff-8c917c681bf8784c25ec13dd769ed68175e6e84897fdcc64dbb20c3ddab7ee5eR38-R44). Reproducing here for convenience:

```kotlin
 private val manager: PanElementStateManager = if (BuildConfig.FLAVOR == "prod") {
        // strictly support only valid Ebt PAN numbers
        PanElementStateManager.forEmptyInput()
    } else {
        // allows whitelist of special Ebt PAN numbers
        PanElementStateManager.DEV_ONLY_forEmptyInput() // will be renamed to NON_PROD_forEmptyInput()
    }
```

devinmorgan: Addressed in #72 to make it easier to see changes!

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
devinmorgan: Addressed in #72 to make it easier to see changes!

dleis612 (approved):

PR #79, opened 2023-08-29
[FX-498] Display element state on screen for qa-tests

dleis612:
Adjust the sample app so that the qa-tests can more easily validate state.


devinmorgan (approved):
Just tested it locally on my machine - amazing addition, Danny! 


In a future PR, we can add red/green color maybe to make extra noise around when things are `true` or `false`

djoksimo (approved):
Nice!

PR #86, opened 2023-09-06
Add details field to ForageErrorObj

devinmorgan:
## What

GoPuff would like the ability to easily read the latest value of the EBT Cash and SNAP values of a PaymentMethod if the PaymentMethod yield an `ebt_error_51` (insufficient funds) error when trying to capture a payment. (see [slack convo](https://joinforage.slack.com/archives/C04FQM5F2DA/p1693922478724489)).

This PR introduces an opaque `.details` field on the `ForageError` that will be `null` for be an instantiation of `ErrorMessageDetails`. Right the only instance is `ErrorMessageDetails.InsufficientFundsDetails` but this can be straightforwardly extended if, in the future, other `ForageError`s ought to carry their own information.

************************************Specifically, this PR does:************************************

- display the full ForageError, including the new details field, as a separate TextView for both
    - SNAP payment capture
    - Cash payment capture
- Add support for ForageErrorObj.details ([linear ticket](https://linear.app/joinforage/issue/FX-528/android-expose-details-field-for-errors)])
- Add helpful comments to error to improve dev experience and reduce integration time

## Why

- Follow new design (strategy B) [described here](https://www.notion.so/Details-in-Error-Object-Discussion-c1948a15620f477ea9e426e60ad701ca?pvs=21)
- Linear ticket: https://linear.app/joinforage/issue/FX-528/android-expose-details-field-for-errors

## Test Plan

- ✅ Android QA Tests pass on Github
- ✅ Unit Tests passed locally

## How

Can be released as-is but should be released at the same times as [the iOS equivalent](https://github.com/teamforage/forage-ios-sdk/pull/101))

Can share the following snippet with GoPuff:

```kotlin
when (response) {
	is ForageApiResponse.Success -> {
	    // handle successful capture
	}
	    val error = response.errors[0]
	    
	    // handle Insufficient Funds error
	    if (error.code == "ebt_error_51") {
	        val details = error.details as ForageErrorDetails.EbtError51Details
	        val (snapBalance, cashBalance) = details

	        // do something with balances ...
	    }
	}
}
```

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
@@ -5,15 +5,29 @@ import org.json.JSONObject
 sealed class ForageApiResponse<out T> {
     data class Success<out T>(val data: T) : ForageApiResponse<T>()
 
-    data class Failure(val errors: List<ForageError>) : ForageApiResponse<Nothing>()
+    data class Failure(val errors: List<ForageError>) : ForageApiResponse<Nothing>() {
+        companion object {
+            fun fromSQSError(sqsError: SQSError): Failure {
+                val forageError = ForageError(
+                    sqsError.statusCode,
+                    sqsError.forageCode,
+                    sqsError.message,
+                    sqsError.details
+                )
+                return Failure(listOf(forageError))
+            }
+        }
+    }
 }
 
 data class ForageError(
     val httpStatusCode: Int,
     val code: String,
-    val message: String
+    val message: String,
+    val details: ErrorMessageDetails? = null
 ) {
     override fun toString(): String {
+        // TODO: should we include the stringified JSON details in the toString() ?
         return "Code: $code\nMessage: $message\nStatus Code: $httpStatusCode"
     }
 }
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
@@ -27,8 +27,7 @@ data class ForageError(
     val details: ErrorMessageDetails? = null
 ) {
     override fun toString(): String {
-        // TODO: should we include the stringified JSON details in the toString() ?
-        return "Code: $code\nMessage: $message\nStatus Code: $httpStatusCode"
+        return "Code: $code\nMessage: $message\nStatus Code: $httpStatusCode\nError Details (below):\n$details"
     }
 }
 
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
@@ -20,10 +20,21 @@ sealed class ForageApiResponse<out T> {
     }
 }
 
+// Learn more about `ForageError`s [here](https://docs.joinforage.app/reference/forage-js-errors#forageerror)
 data class ForageError(
+    // The HTTP status that the Forage API returns in response to the request.
     val httpStatusCode: Int,
+
+    // A short string explaining why the request failed. The [error code](https://docs.joinforage.app/reference/errors#error-codes)
+    // string corresponds to the HTTP status code.
     val code: String,
+
+    // A developer-facing message about the error, not to be displayed to customers.
     val message: String,
+
+    // Additional data associated with certain ForageErrors include for your
+    // convenience. Guaranteed to be present for ForageErrors with details
+    // (e.g. error_code_51 Insufficient Balance). null for all other ForageErrors
     val details: ErrorMessageDetails? = null
 ) {
     override fun toString(): String {
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
@@ -35,7 +35,7 @@ data class ForageError(
     // Additional data associated with certain ForageErrors include for your
     // convenience. Guaranteed to be present for ForageErrors with details
     // (e.g. error_code_51 Insufficient Balance). null for all other ForageErrors
-    val details: ErrorMessageDetails? = null
+    val details: ForageErrorDetails? = null
 ) {
     override fun toString(): String {
         return "Code: $code\nMessage: $message\nStatus Code: $httpStatusCode\nError Details (below):\n$details"
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
@@ -32,7 +32,7 @@ data class ForageError(
     // A developer-facing message about the error, not to be displayed to customers.
     val message: String,
 
-    // Additional data associated with certain ForageErrors include for your
+    // Additional data associated with certain ForageErrors included for your
     // convenience. Guaranteed to be present for ForageErrors with details
     // (e.g. error_code_51 Insufficient Balance). null for all other ForageErrors
     val details: ForageErrorDetails? = null
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/model/Message.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/model/Message.kt
@@ -2,23 +2,45 @@ package com.joinforage.forage.android.network.model
 
 import org.json.JSONObject
 
+sealed class ErrorMessageDetails {
+    data class InsufficientFundsDetails(val snapBalance: String? = null, val cashBalance: String? = null) : ErrorMessageDetails() {
+        companion object {
+            fun from(detailsJson: JSONObject?): InsufficientFundsDetails {
+                // TODO: should probably add a log here if detailsJSON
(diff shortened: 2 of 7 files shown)


devinmorgan:
* main
  * **PR #86** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/86" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/86?utm_source=stack-comment).

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
```diff
@@ -5,15 +5,29 @@ import org.json.JSONObject
...
+        }
+    }
 }
 
 data class ForageError(
     val httpStatusCode: Int,
     val code: String,
-    val message: String
+    val message: String,
+    val details: JSONObject? = null,
```
djoksimo, line 27: Can we add comments for the client, like we did here:

https://github.com/teamforage/forage-ios-sdk/pull/94/files#diff-3aeb1fa61be825a421c835a63db9012dbc1c0eaa859c5e94e23a036813e1f7fdR48-R64

djoksimo, line 27: Can we make the client-facing type stricter than JSONObject? and unmarshall the response into a typed `cashBalance` and `snapBalance`

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
```diff
@@ -5,15 +5,29 @@ import org.json.JSONObject
...
 ) {
     override fun toString(): String {
+        // TODO: should we include the stringified JSON details in the toString() ?
```
djoksimo, line 30: @dleis612 may have additional context, but my guess is that as long as the client doesn't parse the stringified ForageError, it is safe to do this

Would we have to update Android e2e QA tests too?

djoksimo (changes requested):
@devinmorgan thank you for the clean up + tests! I know this is a draft PR, but I'm sharing some comments to expedite the review!

Can you also display the error details in the sample app?

Like [we do here](https://github.com/teamforage/forage-ios-sdk/pull/94/files#diff-b910748c58f44bd34b58c1a1b603b7fa74c27c109bb59fc4146835768bb419d1R151-R160)

And then share a screenshot to showcase that this feature works end-to-end?

devinmorgan:
> @devinmorgan thank you for the clean up + tests! I know this is a draft PR, but I'm sharing some comments to expedite the review!

Exactly what I was hoping for by opening the PR early! Thanks!

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
devinmorgan: yup! done in the latest commit. Once I flushed out the test suite a bit more I started getting really loud errors from just using JSONObject so good call

devinmorgan:
> Can you also display the error details in the sample app?

Done! will update the PR description with screenshots

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
```diff
@@ -5,16 +5,29 @@ import org.json.JSONObject
...
 data class ForageError(
     val httpStatusCode: Int,
     val code: String,
-    val message: String
+    val message: String,
+    val details: ErrorMessageDetails? = null
 ) {
     override fun toString(): String {
-        return "Code: $code\nMessage: $message\nStatus Code: $httpStatusCode"
+        return "Code: $code\nMessage: $message\nStatus Code: $httpStatusCode\nError Details (below):\n$details"
```
devinmorgan, line 30: @dleis612 changed the `.toString()` here just fyi

forage-android/src/main/java/com/joinforage/forage/android/network/model/Message.kt
```diff
@@ -5,25 +5,32 @@ import org.json.JSONObject
...
-    val message: String
+    val message: String,
+    val details: JSONObject? = null,
 ) {
     object SQSErrorMapper {
         fun from(string: String): SQSError {
             val jsonObject = JSONObject(string)
+            println("trying to pint the JSON")
+            println(jsonObject)
+            println(string)
```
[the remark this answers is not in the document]
devinmorgan: done!

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
devinmorgan: yup! Great request

djoksimo (changes requested):
from my understanding (sry if I missed something!):

to access the snap balance and cash balance, the client would have to do type casting:

```kotlin
response.errors[0].details as ErrorMessageDetails.InsufficientFundsDetails).snapBalance
```

which could throw an exception if the response from the Payments API does not perfectly conform to the `ErrorMessageDetails.InsufficientFundsDetails` shape . The client would have to add an additional if statement like this:

```kotlin
val firstError = response.errors.firstOrNull()

if (firstError != null && firstError.details is ErrorMessageDetails.InsufficientFundsDetails) {
    val snapBalance = (firstError.details as ErrorMessageDetails.InsufficientFundsDetails).snapBalance
    // Use snapBalance as needed
} 
```

What are your thoughts on supporting the same interface as the iOS SDK:

```kotlin
val firstError = response.errors.firstOrNull()

firstError?.details?.snapBalance
firstError?.details?.cashBalance
```

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
```diff
@@ -5,16 +5,40 @@ import org.json.JSONObject
...
     val code: String,
-    val message: String
+
+    // A developer-facing message about the error, not to be displayed to customers.
+    val message: String,
+
+    // Additional data associated with certain ForageErrors include for your
+    // convenience. Guaranteed to be present for ForageErrors with details
+    // (e.g. error_code_51 Insufficient Balance). null for all other ForageErrors
+    val details: ErrorMessageDetails? = null
```
djoksimo, line 38: can we call this `ForageErrorDetails` ?

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
djoksimo: If this is causing issues, can we revert this line to the original `toString()` and display the insufficient funds details separately in the sample app?

djoksimo (changes requested):

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
devinmorgan: Nice call. Making a separate <TextView /> to display the `snapResponseError` and the `cashResponseError` sounds like the way to go 👍. With any luck the qa-tests will be resolved now:
1. First issue that caused the QA-tests to fail was that I displayed `forageError.toString()` instead of `forageError.message` (mobile-qa-tests expect that later)
2. Second issue that caused the QA-tests to fail was wrapping the PaymentCapture view in a Scroll view since the Cash payment get's cut off. I'm putting [this scroll-view work](https://linear.app/joinforage/issue/FX-548/android-update-payment-capture-view-to-have-scroll-view) to clean up since it will require a PR to the mobile-qa-tests which doesn't feel like a priority right now

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
devinmorgan: I'm commenting now that I've had to investigate QA tests failing a few times. Updating the `.toString()` does not break the QA tests directly. The android mobile QA tests assert against `error.message` (see [here](https://github.com/teamforage/forage-android-sdk/blob/b62429777b2273eccb7509ca48fe924e3bed28e4/sample-app/src/main/java/com/joinforage/android/example/ui/complete/flow/payment/capture/FlowCapturePaymentViewModel.kt#L101) and [here](https://github.com/search?q=repo%3Ateamforage/mobile-qa-tests%20error_clauses&type=code)). This PR aims to display the full error (including the new `details`) field by introducing a new `<TextView />` that displays the whole thing and leaving the original `snapResponse` and `cashResponse` `<TextView / >`s untouched. Lmk if that clarifies!

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
```diff
@@ -5,16 +5,40 @@ import org.json.JSONObject
...
+
+    // A short string explaining why the request failed. The [error code](https://docs.joinforage.app/reference/errors#error-codes)
+    // string corresponds to the HTTP status code.
     val code: String,
-    val message: String
+
+    // A developer-facing message about the error, not to be displayed to customers.
+    val message: String,
+
+    // Additional data associated with certain ForageErrors include for your
```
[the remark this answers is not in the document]
devinmorgan: fixed !

djoksimo (approved):
Nice work!

PR #89, opened 2023-09-11
[FX-385] Add trace ID to Android SDK, naive prefixes to logs, and data to VGS calls

dleis612:
## What
Added trace IDs to the Android SDK and improved the current log schema with prefixes. The prefixes are naive at this point and should be adjusted to comply with DRY. Also, I added some extra pieces of data to the VGS logs.

## Test Plan
Tested locally and verified in Datadog.

## Demo
Screenshot 2023-09-12 at 10 18 14 AM


devinmorgan (approved):
Nice with the traceIDs!

PR #94, opened 2023-09-12
[POP-180] Adding jitter to retry logic

nitinsampath:
## What

Adding +/- 25ms jitter to polling so that we can avoid a thundering herd issue if our clients retry all at the same time. Now all retries to polling will have some randomness added to their delay interval.

## Test Plan

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/Utils.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/Utils.kt
@@ -1,0 +1,14 @@
+package com.joinforage.forage.android
+
+import kotlin.random.Random
+
+/**
+ * We generate a random jitter amount to add to our retry delay when polling for the status of
+ * Payments and Payment Methods so that we can avoid a thundering herd scenario in which there are
+ * several requests retrying at the same exact time.
+ *
+ * Returns a random integer between -25 and 25
+ */
+internal fun getJitterAmount(random: Random = Random.Default): Int {
+    return random.nextInt(-25, 26)
+}
--- a/forage-android/src/test/java/com/joinforage/forage/android/utils/UtilsTest.kt
+++ b/forage-android/src/test/java/com/joinforage/forage/android/utils/UtilsTest.kt
@@ -1,0 +1,24 @@
+package com.joinforage.forage.android.utils
+
+import com.joinforage.forage.android.getJitterAmount
+import kotlinx.coroutines.test.runTest
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import kotlin.random.Random
+
+class UtilsTest {
+    @Test
+    fun `Jitter value should be in range -25 to 25`() = runTest {
+        val fixedRandomZero = Random(13)
+        val jitterZero = getJitterAmount(fixedRandomZero)
+        assertThat(jitterZero).isEqualTo(0)
+
+        val fixedRandomMin = Random(16)
+        val jitterMin = getJitterAmount(fixedRandomMin)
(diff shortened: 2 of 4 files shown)


forage-android/src/test/java/com/joinforage/forage/android/utils/UtilsTest.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class UtilsTest {
+    @Test
+    fun `Jitter value should be in range -25 to 25`() = runTest {
+        val jitter = generateJitter()
+        assertThat(jitter).isGreaterThanOrEqualTo(-25)
+        assertThat(jitter).isLessThanOrEqualTo(25)
+    }
```
djoksimo, line 14: this introduces randomness to our unit tests and may not catch issues with the `generateJitter`. Can we make this test deterministic?

```kotlin
internal fun generateJitter(random: Random = Random.Default): Int {
    return random.nextInt(-25, 26)
}

val fixedRandom = Random(12345) // Update seed value here
val jitter = generateJitter(fixedRandom)
```

forage-android/src/main/java/com/joinforage/forage/android/Utils.kt
```diff
@@ -0,0 +1,8 @@
+package com.joinforage.forage.android
+
+import kotlin.random.Random
+
+fun generateJitter(): Int {
```
djoksimo, line 5: can we mark this as an `internal fun` to avoid exposing this to the client?

djoksimo: Ik i mentioned generateJitter in slack (sry), can we rename this to

```kotlin
fun getJitterAmount
```

and maybe add a comment explaining why we add jitter for maintainability?

djoksimo (changes requested):
Thank you @nitinsampath !

Left a few comments

forage-android/src/test/java/com/joinforage/forage/android/utils/UtilsTest.kt
nitinsampath: I'm going to seed the RNG at the method level so it will look something like:
```
internal fun getJitterAmount(): Int {
    val random = Random(999)
    return random.nextInt(-25, 26)
}
```

forage-android/src/main/java/com/joinforage/forage/android/Utils.kt
nitinsampath: [c5943c9](https://github.com/teamforage/forage-android-sdk/pull/94/commits/c5943c94e45dd5b967eff50e4cf70593a4c3eac7)

forage-android/src/test/java/com/joinforage/forage/android/utils/UtilsTest.kt
nitinsampath: [c5943c9](https://github.com/teamforage/forage-android-sdk/pull/94/commits/c5943c94e45dd5b967eff50e4cf70593a4c3eac7)

djoksimo: fixing the seed at the method level will make the function return a fixed value each time.

As seen in this sandbox: https://pl.kotl.in/gX-Sm2V5A

Can we use `Random.Default` for the actual method and the fixed seed value for the test?

djoksimo (changes requested):

djoksimo (approved):
Nice 🚀 🚀

PR #95, opened 2023-09-13
[FX-395]  Add latency vault type response code to dd logs

dleis612:
## What
Add latency metrics to log statements for the vaults.

## Test Plan

- ❌ The reviewer should manually test the changes in this PR.

## Demo
Screenshot 2023-09-15 at 12 44 13 PM


forage-android/src/main/java/com/joinforage/forage/android/core/telemetry/Metrics.kt
```diff
@@ -0,0 +1,204 @@
...
+    the user. Ex of a balance action:
+    Timer Begins -> [GET] EncryptionKey -> [GET] PaymentMethod -> [POST] to VGS/BT ->
+    [GET] Poll for Response -> [GET] PaymentMethod -> Timer Ends -> Return Balance
+     */
+    CUSTOMER_PERCEIVED_RESPONSE_TIME("customer_perceived_response_time");
+
+    override fun toString(): String {
+        return value
+    }
+}
```
djoksimo, line 45: the vault log and the customer-perceived log will/can be re-used for tracking error rates (like I did for iOS SDK), meaning that these logs are not time-specific, so the `_time` suffix can be removed

can we do something like 

```kotlin
internal enum class EventName {
  VAULT_RESPONSE
  
  CUSTOMER_PERCEIVED_RESPONSE
}
```

@devinmorgan @miketikh

forage-android/src/main/java/com/joinforage/forage/android/core/telemetry/Metrics.kt
```diff
@@ -0,0 +1,204 @@
...
+    override fun logWithResponseAttributes(
+        metricsLogger: Log?,
+        responseAttributes: Map<String, Any>
+    ) {
+        val responseTime = responseAttributes[MetricsConstants.RESPONSE_TIME_MS]
+
+        if (responseTime == null) {
+            metricsLogger?.e("[Metrics] Incomplete or missing response attributes. Could not log metric.")
+            return
+        }
```
djoksimo, line 189: we should assert that `EventName` (or RESPONSE_TYPE) and `action` also set to the right EventName here

to ensure our metrics are tuned correctly and to help us distinguish between BIs + captures

```kotlin
if (responseTime == null || eventName != CUSTOMER_PERCEIVED_RESPONSE || action == null) {
...
}
```

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -160,6 +180,12 @@ internal class CapturePaymentRepository(
             )
         )
 
+        // END METRICS TRACKING UPON SUCCESSFUL FULL TRIP
+        // ------------------------------------------------------
+        responseMonitor.end()
+        responseMonitor.logResult()
+        // ------------------------------------------------------
+
         return paymentService.getPayment(paymentRef = paymentRef)
```
djoksimo, line 189: there is a `getPayment` request on line 189 after the call to `logResult`, so the customer-perceived response metric may not be accurate and include the full capture path.

Should we call end at the entry `ForageSDK.kt` level and set a boundary around the beginning of the `checkBalance` method and the time when the client gets a response.

In this case, we also wouldn't need to worry about drilling monitors through numerous request methods

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -66,8 +84,8 @@ internal class CapturePaymentRepository(
     private suspend fun collectPinToCapturePayment(
         paymentRef: String,
         cardToken: String,
-        encryptionKey: String
-
+        encryptionKey: String,
+        responseMonitor: ResponseMonitor
```
djoksimo, line 88: **Can ignore this comment if we purse the `ForageSDK.kt` approach described in my other comment**

> Should we call end at the entry ForageSDK.kt level and set a boundary around the beginning of the checkBalance method and the time when the client gets a response.

----
to avoid drilling the `responseMonitor` through every request method

could we keep track of `responseMonitor` in a map?

```kotlin
internal object MeasurementStore {
  private val store: HashMap<EventName, ResponseMonitor> = HashMap()

  fun set(key: EventName, instance: ResponseMonitor) {
     store[key] = instance
  }

  fun get(key: EventName): ResponseMonitor? {
    return store[key]
  }
}

customerPerceivedMonitor.start()
MeasurementStore.set(CUSTOMER_PERCEIVED, customerPerceivedMonitor)


val customerPerceivedMonitor = MeasurementStore.get(CUSTOMER_PERCEIVED)
customerPerceivedMonitor.end()
```

@devinmorgan would also like to get your thoughts on alternatives to parameter drilling the responseMonitor.

forage-android/src/test/java/com/joinforage/forage/android/network/data/TestPinCollector.kt
```diff
@@ -58,6 +59,10 @@ internal class TestPinCollector : PinCollector {
         return token
     }
 
+    override fun getVaultType(): VaultType {
+        return VaultType.VGS_VAULT_TYPE
+    }
```
djoksimo, line 64: nice!

forage-android/src/test/java/com/joinforage/forage/android/core/telemetry/metrics/ResponseMonitorTest.kt
djoksimo, line 1: thank you for adding these awesome tests!

forage-android/src/main/java/com/joinforage/forage/android/core/telemetry/Metrics.kt
dleis612: `vaultAction` and `vaultType` can't be null and we can't validate their value at this point.

djoksimo: sounds good, i've amended my comment^

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -120,9 +174,12 @@ internal class BTPinCollector(
                 "payment_ref" to paymentRef
             )
         )
+
+        measurement.start()
         val response = runCatching {
             bt.proxy.post(proxyRequest)
```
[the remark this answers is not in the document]
djoksimo: Alternatively, instead of relying on HTTP status codes for tracking error rates, we could use our [error codes](https://docs.joinforage.app/reference/errors#sdk-errors) to distinguish between errors from Forage vs. from Basis Theory.
 
IN Datadog: filter out error codes like ebt_error_51 (insufficient funds), ebt_error_55 from the Vault Proxy error rate monitors + graphs.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -134,10 +148,21 @@ class ForageSDK : ForageSDKInterface {
...
-            logger = logger
+            logger = logger,
+            responseMonitor = measurement
         ).checkBalance(
             paymentMethodRef = paymentMethodRef
         )
+        measurement.end()
+
+        val outcome = if (response is ForageApiResponse.Failure) {
+            EventOutcome.FAILURE
```
miketikh, line 159: Just reading through the comments to get up to speed, but did we also want to add the `forage_error_code` here?

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -64,21 +96,23 @@ internal class BTPinCollector(
...
                 )
-            } catch (e: JSONException) { }
+            } catch (_: JSONException) { }
             logger.i(
                 "[BT] Received successful response from BasisTheory",
                 attributes = mapOf(
                     "merchant_ref" to merchantAccount,
                     "payment_method_ref" to paymentMethodRef
                 )
             )
```
miketikh, line 112: this logic's a bit confusing for me. So if I understand, we're:
- checking if the response was successful
  - if it was successful, try getting the error in a try block, which throws an error if not available
  - catch the thrown error from when error isn't available, and use that to determine a successful response
  
If that's correct, could it be more intuitive to do it the same way we do in the js repo?
```
   val forageResponse = response.getOrNull()
            try {
                // This will return null if an error isn't available, instead of throwing
                val forageApiError = ForageApiError.ForageApiErrorMapper.from(forageResponse.toString())
                
                if (forageApiError) { // handleError}
                // Else handle success?
                }
```

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
dleis612: So I actually was setting the value within the function call above... but this is a better spot! I'm going to update.

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
dleis612: I'm going to make a new ticket to attempt to address this because it is a little weird as of now.

djoksimo (approved):

PR #96, opened 2023-09-18
[FX-523] Fix height and width attribute naming convention

dleis612:
## What
Fix the naming convention of publicly accessible fields.

## Test Plan

- ❌ I've added unit tests for this change.
- ❌ The reviewer should manually test the changes in this PR.


djoksimo (approved):
Nice catch! Thank you

PR #101, opened 2023-09-20
[POP-76] Configure polling with feature flag

nitinsampath:
## What

Reading the `iso-polling-wait-intervals` feature flag value to determine how long to wait between each polling retry.

## Why

Make polling more configurable instead of waiting 1 second every poll attempt. Leads to faster response times from our clients.

## Test Plan

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
@@ -5,6 +5,7 @@ import com.joinforage.forage.android.core.StopgapGlobalState
 import com.joinforage.forage.android.core.telemetry.Log
 import com.launchdarkly.sdk.ContextKind
 import com.launchdarkly.sdk.LDContext
+import com.launchdarkly.sdk.LDValue
 import com.launchdarkly.sdk.android.LDClient
 import com.launchdarkly.sdk.android.LDConfig
 import com.launchdarkly.sdk.android.integrations.TestData
@@ -20,6 +21,7 @@ internal enum class VaultType(val value: String) {
 
 internal object LDFlags {
     const val VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG = "vault-primary-traffic-percentage"
+    const val ISO_POLLING_WAIT_INTERVALS = "iso-polling-wait-intervals"
 }
 
 internal object LDContexts {
@@ -33,7 +35,9 @@ internal object LDContextKind {
 internal object LDManager {
     private val LD_MOBILE_KEY = StopgapGlobalState.envConfig.ldMobileKey
     private var internalVaultType: VaultType? = null
+
     private var internalLogger: Log = Log.getSilentInstance()
+    private var client: LDClient? = null
 
     internal var vaultType: VaultType?
         get() = internalVaultType
@@ -51,15 +55,7 @@ internal object LDManager {
             }
         }
 
-    // vaultType is instantiated lazily and is a singleton. Once we set the vault type once, we don't
-    // want to overwrite it! We must take in the application as a parameter, which means that a
-    // ForagePINEditText must be rendered before any of the ForageSDKApi functions are called.
-    internal fun getVaultProvider(app: Application, logger: Log, dataSource: TestData? = null): VaultType {
-        if (vaultType != null) {
-            return vaultType as VaultType
-        }
-        internalLogger = logger
-        // Datasource is required for testing purposes!
+    internal fun initialize(app: Application, logger: Log, dataSource: TestData? = null) {
         val ldConfig = if (dataSource != null) {
             LDConfig.Builder()
                 .mobileKey(LD_MOBILE_KEY)
@@ -72,11 +68,25 @@ internal object LDManager {
         }
         val contextKind = ContextKind.of(LDContextKind.SERVICE)
         val context = LDContext.create(contextKind, LDContexts.ANDROID_CONTEXT)
-        val client = LDClient.init(app, ldConfig, context, 0)
+        client = LDClient.init(app, ldConfig, context, 0)
+
+        internalLogger = logger
+    }
+
+    // vaultType is instantiated lazily and is a singleton. Once we set the vault type once, we don't
+    // want to overwrite it! We must take in the application as a parameter, which means that a
+    // ForagePINEditText must be rendered before any of the ForageSDKApi functions are called.
+    internal fun getVaultProvider(): VaultType {
+        if (vaultType != null) {
+            return vaultType as VaultType
+        }
         // default to 100% VGS usage in case LD flag retrieval fails
-        val vaultPercent = client.doubleVariation(LDFlags.VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG, 0.0)
+        val defaultVal = 0.0
+        val vaultPercent =
+            client?.doubleVariation(LDFlags.VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG, defaultVal) ?: defaultVal
         internalLogger.i("[LaunchDarkly] Vault percent of $vaultPercent return from LD")
         val randomNum = Math.random() * 100
+
         vaultType = if (randomNum < vaultPercent) {
             VaultType.BT_VAULT_TYPE
         } else {
@@ -84,5 +94,22 @@ internal object LDManager {
         }
         internalLogger.i("[LaunchDarkly] Vault type set to $vaultType")
         return vaultType as VaultType
+    }
+
+    internal fun getPollingIntervals(): LongArray {
+        val defaultVal = LDValue.buildObject().put(
+            "intervals",
+            LDValue.Convert.Long.arrayFrom(List(10) { 1000L })
+        ).build()
+
+        val jsonIntervals = client?.jsonValueVariation(LDFlags.ISO_POLLING_WAIT_INTERVALS, defaultVal) ?: defaultVal
+
+        val intervals = jsonIntervals.get("intervals")
+
+        // Converting the LDArray into a LongArray
+        val pollingList = LongArray(intervals.size()) { intervals.get(it).longValue() }
+
+        internalLogger.i("[LaunchDarkly] polling intervals $pollingList")
+        return pollingList
     }
 }
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
@@ -1,5 +1,6 @@
 package com.joinforage.forage.android.network.data
 
+import com.joinforage.forage.android.LDManager
 import com.joinforage.forage.android.collect.PinCollector
 import com.joinforage.forage.android.core.telemetry.Log
 import com.joinforage.forage.android.getJitterAmount
@@ -91,6 +92,7 @@ internal class CapturePaymentRepository(
         paymentRef: String
     ): ForageApiResponse<String> {
         var attempt = 1
+        val pollingIntervals = LDManager.getPollingIntervals()
 
         while (true) {
             logger.i(
@@ -151,8 +153,15 @@ internal class CapturePaymentRepository(
                 return ForageApiResponse.Failure(listOf(ForageError(500, "unknown_server_error", "Unknown Server Error")))
             }
 
+            val index = attempt - 1
+            var intervalTime: Long = if (index < pollingIntervals.size) {
+                pollingIntervals[index]
+            } else {
+                1000L
+            }
+
             attempt += 1
-            delay(POLLING_INTERVAL_IN_MILLIS + getJitterAmount())
+            delay(intervalTime + getJitterAmount())
         }
 
         logger.i(
@@ -167,7 +176,6 @@ internal class CapturePaymentRepository(
     }
 
     companion object {
-        private const val POLLING_INTERVAL_IN_MILLIS = 1000L
         private const val MAX_ATTEMPTS = 10
 
         private fun ForageApiResponse<String>.getStringResponse() = when (this) {
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/data/CheckBalanceRepository.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/data/CheckBalanceRepository.kt
@@ -1,5 +1,6 @@
 package com.joinforage.forage.android.network.data
 
+import com.joinforage.forage.android.LDManager
 import com.joinforage.forage.android.collect.PinCollector
 import com.joinforage.forage.android.core.telemetry.Log
 import com.joinforage.forage.android.getJitterAmount
@@ -74,6 +75,7 @@ internal class CheckBalanceRepository(
         paymentMethodRef: String
     ): ForageApiResponse<String> {
         var attempt = 1
+        val pollingIntervals = LDManager.getPollingIntervals()
 
         while (true) {
             logger.i(
@@ -134,8 +136,15 @@ internal class CheckBalanceRepository(
                 return ForageApiResponse.Failure(listOf(ForageError(500, "unknown_server_error", "Unknown Server Error")))
             }
 
+            val index = attempt - 1
+            var intervalTime: Long = if (index < pollingIntervals.size) {
+                pollingIntervals[index]
+            } else {
+                1000L
+            }
+
             attempt += 1
-            delay(POLLING_INTERVAL_IN_MILLIS + getJitterAmount())
+            delay(intervalTime + getJitterAmount())
         }
 
         logger.i(
@@ -163,7 +172,6 @@ internal class CheckBalanceRepository(
     }
 
     companion object {
-        private const val POLLING_INTERVAL_IN_MILLIS = 1000L
         private const val MAX_ATTEMPTS = 10
 
         private fun ForageApiResponse<String>.getStringResponse() = when (this) {
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
@@ -102,7 +102,10 @@ class ForagePINEditText @JvmOverloads constructor(
         // its relying on forageConfig under the hood. This
         // relationship will be made explicit once we drop
         // StopgapGlobalState
-        val vaultType = LDManager.getVaultProvider(context.applicationContext as Application, logger)
+
+        LDManager.initialize(context.applicationContext as Application, logger)
+        val vaultType = LDManager.getVaultProvider()
+
         _SET_ONLY_vault = if (vaultType == VaultType.BT_VAULT_TYPE) {
             btVaultWrapper
         } else {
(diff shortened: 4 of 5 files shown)


forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -91,7 +92,8 @@ internal class CapturePaymentRepository(
         contentId: String,
         paymentRef: String
     ): ForageApiResponse<String> {
-        var attempt = 1
+        var attempt = 0
+        val pollingIntervals = LDManager.getPollingIntervals() // TODO how to get the Application and logger params
```
nitinsampath, line 96: how do I get `Application` and `logger` as params here

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -80,4 +99,32 @@ internal object LDManager {
...
+        } else {
+            LDConfig.Builder()
+                .mobileKey(LD_MOBILE_KEY)
+                .build()
+        }
+        val contextKind = ContextKind.of(LDContextKind.SERVICE)
+        val context = LDContext.create(contextKind, LDContexts.ANDROID_CONTEXT)
+        val client = LDClient.init(app, ldConfig, context, 0)
+
+        val jsonIntervals = client.jsonValueVariation(LDFlags.ISO_POLLING_WAIT_INTERVALS, LDValue.buildObject().put("intervals", LDValue.buildArray().add(1000).build()).build())
```
nitinsampath, line 123: [LD](https://docs.launchdarkly.com/sdk/client-side/android/migration-2-to-3/?q=jsonvaluevariation#understanding-changes-to-the-variation-methods) introduced `LDValue` in the `jsonValueVariation` method.

We need to determine how to use `LDValue` properly here.

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -86,12 +79,25 @@ internal object LDManager {
...
+            return vaultType as String
+        }
         // default to 100% VGS usage in case LD flag retrieval fails
-        val vaultPercent = client.doubleVariation(LDFlags.VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG, 0.0)
+        val vaultPercent =
+            client?.doubleVariation(LDFlags.VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG, 0.0)
         internalLogger.i("[LaunchDarkly] Vault percent of $vaultPercent return from LD")
         val randomNum = Math.random() * 100
-        vaultType = if (randomNum < vaultPercent) {
+        vaultType = if (randomNum < vaultPercent!!) {
```
djoksimo, line 100: doing a strict null assertion on this can cause a crash if the request to LD fails. Since we can set fallbacks, can we fallback to the default 0 value here?

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -72,17 +85,54 @@ internal object LDManager {
...
     }
+
+    internal fun getPollingIntervalsFeatureFlag(): IntArray {
+        if (pollingIntervals != null) {
+            return pollingIntervals as IntArray
+        }
+
+        var pollingList = IntArray(0)
+
+        val jsonIntervals =
```
dleis612, line 122: Could you do the same thing for this setter by including the default value here? Also, can you set the default value as a variable above for readability?

If the LDClient is null, I think this should work right?

```kotlin
val defaultVal = LDValue.buildObject().put("intervals", LDValue.buildArray().add(1000).build()).build()
val jsonIntervals = client?.jsonValueVariation(LDFlags.ISO_POLLING_WAIT_INTERVALS, defaultVal) ?: defaultVal
```

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -72,17 +85,54 @@ internal object LDManager {
...
+            client?.jsonValueVariation(LDFlags.ISO_POLLING_WAIT_INTERVALS, LDValue.buildObject().put("intervals", LDValue.buildArray().add(1000).build()).build())
```
dleis612, line 123: What will be the behavior of the default value if it's the array `[1000]`? Are we only going to poll once after 1 second?

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -90,7 +91,8 @@ internal class CapturePaymentRepository(
         contentId: String,
         paymentRef: String
     ): ForageApiResponse<String> {
-        var attempt = 1
+        var attempt = 0
+        val pollingIntervals = LDManager.getPollingIntervalsFeatureFlag() // TODO how to get the Application and logger params
```
dleis612, line 95: Don't believe you'll need to get the Application or logger params here anymore.

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -151,8 +153,14 @@ internal class CapturePaymentRepository(
                 return ForageApiResponse.Failure(listOf(ForageError(500, "unknown_server_error", "Unknown Server Error")))
             }
 
+            var intervalTime: Long = if (pollingIntervals != null && attempt < pollingIntervals.size) {
+                pollingIntervals[attempt].toLong()
+            } else {
+                1000L
```
dleis612, line 159: Ah I see you default to polling for every second if there are issues with the `pollingIntervals` value.

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
nitinsampath: forgot to delete that comment, whoops

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -72,17 +85,54 @@ internal object LDManager {
...
             VaultType.BT_VAULT_TYPE
         } else {
             VaultType.VGS_VAULT_TYPE
         }
         internalLogger.i("[LaunchDarkly] Vault type set to $vaultType")
         return vaultType as VaultType
     }
+
+    internal fun getPollingIntervalsFeatureFlag(): IntArray {
+        if (pollingIntervals != null) {
```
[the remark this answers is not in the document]
nitinsampath: Oh i thought this was the pattern for all feature flags in general. I can just return the latest polling intervals every time instead of storing the value in this singleton pattern

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
nitinsampath: yeah that works, ill make the change

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
nitinsampath: I just made the default LDValue a list of size 1 because it seems so verbose to have to do `.add(1000)` 10 times. Lemme see if there's an easier way to define a list rather than doing something like `.add(1000).add(1000).add(1000).add(1000).add(1000).add(1000).add(1000)`

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
nitinsampath: yes, we'll fall back to 1 second if `pollingIntervals` was not populated correctly or if the length of the array is less than the number of max attempts.

If length of pollingIntervals is 8, but max attempts is 10, the last 2 attempts would wait 1 second each before retrying

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
nitinsampath: I was able to get it to work by doing 
```
        val defaultVal = LDValue.buildObject().put("intervals",
            LDValue.Convert.Integer.arrayFrom(listOf(1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000))).build()
```
This way it is more explicit what the default value is

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -33,7 +35,9 @@ internal object LDContextKind {
 internal object LDManager {
     private val LD_MOBILE_KEY = StopgapGlobalState.envConfig.ldMobileKey
     private var internalVaultType: VaultType? = null
+
     private var internalLogger: Log = Log.getSilentInstance()
+    var client: LDClient? = null
```
dleis612, line 40: Make private por favor

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -85,4 +95,26 @@ internal object LDManager {
         internalLogger.i("[LaunchDarkly] Vault type set to $vaultType")
         return vaultType as VaultType
     }
+
+    internal fun getPollingIntervalsFeatureFlag(): IntArray {
```
dleis612, line 99: nit: Can you remove the "FeatureFlag" portion of this function name so it's the same style as `getVaultProvider`?

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -85,4 +95,26 @@ internal object LDManager {
...
+        ).build()
+
+        val jsonIntervals = client?.jsonValueVariation(LDFlags.ISO_POLLING_WAIT_INTERVALS, defaultVal) ?: defaultVal
+
+        val intervals = jsonIntervals.get("intervals")
+
+        // Converting the LDArray into an IntArray
+        // If we cannot access LD, return int array [1000] as a default
+        for (i in 0 until intervals.size()) {
+            pollingList += intervals?.get(i)?.intValue() ?: 1000
```
dleis612, line 114: Should be able to remove the `?: 1000` portion now right? `jsonIntervals` always has a non-null value.

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -151,8 +153,14 @@ internal class CapturePaymentRepository(
                 return ForageApiResponse.Failure(listOf(ForageError(500, "unknown_server_error", "Unknown Server Error")))
             }
 
+            var intervalTime: Long = if (pollingIntervals != null && attempt < pollingIntervals.size) {
```
dleis612, line 156: `pollingIntervals` should be non-null now.

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -51,15 +55,7 @@ internal object LDManager {
...
-    // vaultType is instantiated lazily and is a singleton. Once we set the vault type once, we don't
-    // want to overwrite it! We must take in the application as a parameter, which means that a
-    // ForagePINEditText must be rendered before any of the ForageSDKApi functions are called.
-    internal fun getVaultProvider(app: Application, logger: Log, dataSource: TestData? = null): VaultType {
-        if (vaultType != null) {
-            return vaultType as VaultType
-        }
-        internalLogger = logger
-        // Datasource is required for testing purposes!
+    internal fun setClientAndLogger(app: Application, logger: Log, dataSource: TestData? = null) {
```
djoksimo, line 58:
```suggestion
    internal fun initialize(app: Application, logger: Log, dataSource: TestData? = null) {
```

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -85,4 +95,26 @@ internal object LDManager {
         internalLogger.i("[LaunchDarkly] Vault type set to $vaultType")
         return vaultType as VaultType
     }
+
+    internal fun getPollingIntervals(): IntArray {
```
djoksimo, line 99: why do we return `IntArray` and then convert to long after getting the polling intervals?

Could we return a `LongArray`

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -85,4 +95,26 @@ internal object LDManager {
...
+        var pollingList = IntArray(0)
+
+        val defaultVal = LDValue.buildObject().put(
+            "intervals",
+            LDValue.Convert.Integer.arrayFrom(listOf(1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000, 1000))
```
djoksimo, line 104:
```suggestion
            LDValue.Convert.Integer.arrayFrom(List(10) { 1000 })
```

could this work^

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -85,4 +95,26 @@ internal object LDManager {
...
+
+        val jsonIntervals = client?.jsonValueVariation(LDFlags.ISO_POLLING_WAIT_INTERVALS, defaultVal) ?: defaultVal
+
+        val intervals = jsonIntervals.get("intervals")
+
+        // Converting the LDArray into an IntArray
+        // If we cannot access LD, return int array [1000] as a default
+        for (i in 0 until intervals.size()) {
+            pollingList += intervals.get(i).intValue()
+        }
```
djoksimo, line 115:
```suggestion
      val pollingList = IntArray(intervals.size()) { intervals.get(it).intValue() }
```

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -151,8 +153,14 @@ internal class CapturePaymentRepository(
...
+            var intervalTime: Long = if (attempt < pollingIntervals.size) {
+                pollingIntervals[attempt].toLong()
+            } else {
+                1000L
+            }
+
             attempt += 1
-            delay(POLLING_INTERVAL_IN_MILLIS + getJitterAmount())
+            delay(intervalTime + getJitterAmount())
         }
```
djoksimo, line 164: the logic in this method seems almost the same (besides the log statements) as the logic in the balance polling. I've created a ticket to consolidate the polling logic into a reusable method/class to reduce this duplication.

https://linear.app/joinforage/issue/FX-610/consolidate-dedupe-polling-logic-in-android

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
nitinsampath: Yes, then we can get rid of the `toLong()` conversion when reading the value from the list

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
nitinsampath: Thanks for making this - i'll keep that change separate from this PR actually

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
nitinsampath: TIL lambda expressions exist in Kotlin!

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -90,7 +91,8 @@ internal class CapturePaymentRepository(
         contentId: String,
         paymentRef: String
     ): ForageApiResponse<String> {
-        var attempt = 1
+        var attempt = 0
```
devinmorgan, line 94: technically, if you change `attempt` from `1` -> `0`, the if-statement `if (attempt == MAX_ATTEMPTS) { ... }` will get triggered after `MAX_ATTEMPTS + 1`. Is this OK? If we are making any gaurantees to customers that a timeout will happen in `pollingInterval * MAX_ATTEMPTS` or sooner than we may want to change the if-statement below to be `if (attempt == MAX_ATTEMPTS - 1) { ... }`.

IMO I think it would be easier to read if the if-statement remained `if (attempt == MAX_ATTEMPTS) { ... }` and we updated this code:
```kotlin
var intervalTime: Long = if (attempt < pollingIntervals.size) {
    pollingIntervals[attempt]
} else { ... }
```

to be 

```kotlin
val index = attempt - 1
var intervalTime: Long = if (index < pollingIntervals.size) {
    pollingIntervals[index]
} else { ... }
```

forage-android/src/main/java/com/joinforage/forage/android/network/data/CheckBalanceRepository.kt
```diff
@@ -134,8 +136,14 @@ internal class CheckBalanceRepository(
                 return ForageApiResponse.Failure(listOf(ForageError(500, "unknown_server_error", "Unknown Server Error")))
             }
 
+            var intervalTime: Long = if (attempt < pollingIntervals.size) {
+                pollingIntervals[attempt]
+            } else {
```
devinmorgan, line 141: Same remark as above but to this part of the code as well
```suggestion
val index = attempt - 1
var intervalTime: Long = if (index < pollingIntervals.size) {
    pollingIntervals[index]
} else {
```

devinmorgan (changes requested):
Literally just the one comment (in two places). Otherwise PR looks good - appreciate all of your work on this feature, Nitin!!

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
nitinsampath: That makes sense, thanks for the good catch! I will use the `var index = attempt - 1` approach

devinmorgan (approved):
Nice work man! 🚢 it!

djoksimo (approved):
Code LGTM; nice work @nitinsampath !  

Before we release this in the next version:

Can we set the feature flag against [the staging environment and android-sdk-service](https://app.launchdarkly.com/default/staging/features/iso-polling-wait-intervals/targeting) to make sure this is working reliably end-to-end?

I'm curious about edge case test scenarios to make sure this works reliably: 

1. When the feature rule is turned off, do balance checks / QA tests succeed using the default rule?

2. When the intervals are short or depleted; will the polling interval fallback to 1000ms? [We can look at logs in datadog with timestamps](https://app.datadoghq.com/logs?query=service%3Aandroid-sdk%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=sliding&stream_sort=desc&viz=stream&from_ts=1696371835209&to_ts=1696372735209&live=true) to ensure this is working reliably.

```json
{ "intervals": [1] }
```

3. When the "intervals" value is invalid (with 4-eyes policy and pre-defined options, this is unlikely to happen).  Maybe something worth considering? Will launch darkly just use the default value or will balance checks cause crashes?

```json
{ "intervals": ["oops-invalid"] }

// typo
{ "interval$":  [1000] }
```

PR #102, opened 2023-09-20
Milestone 2 - Introduce `StopgapGlobalState`

devinmorgan:
## What
This is the PR for [Milestone 2](https://www.notion.so/joinforage/Plan-for-On-Time-Delivery-of-Android-3-0-Breaking-Changes-226d8ee6f8294d2694b1bb451791960b?pvs=4#f7302e2ba6df42fc80cd67c6e36f1fa9) of the plan to ship Android v3 to GoPuff by tomorrow EOD

## Why
We will need to temporarily rely on global state (via singleton) to ship Android v3 public contract changes ASAP without getting to [all of the required internal changes](https://www.notion.so/joinforage/Android-3-0-Design-Proposal-1ee6be95544d4dc78e7a7e7c88f46bbc?pvs=4#deb5c9f46bf74da9abc18f0b5f5ea4f4).

## Test Plan
- ✅ I've added unit tests for this change. Specifically, I've added tests for the parsing session token logic
- ❌ The reviewer should NOT manually test the changes in this PR. This PR does not change any observable behavior. Those changes will be in the next PR in this stack

## How
Wait to merge this one until the next PR in this stack is ready to be merged

PR #98, opened 2023-09-20
Drop canTokenizePanElementValue

devinmorgan:
## What
There was an unused method in the `PanElementStateManager` class that was breaking the build pipeline seemingly after the introduction of Android v3 changes.

## Why

I opted to separate this change into its own PR to make reviewing the previous PR easier - I figured this would be an easier story to consume.

## How
Can be merged down once approved, but I would wait until #97 is approved before merging this into #97 to keep everything easy to read. Just my two cents

PR #97, opened 2023-09-20
[BREAKING] Milestone 1 - Add Proposed Android v3 Interfaces

devinmorgan:
## What
This is the first of 3 milestones to ship the Android v3 public contract changes to GoPuff by Thursday 9/21. As dicussused [here](https://www.notion.so/joinforage/Android-3-0-Design-Proposal-1ee6be95544d4dc78e7a7e7c88f46bbc?pvs=4)

This first milestone:
1. introduces
  1. `ForageContext` data class
  2. `TokenizeEBTCardParams` data class
  3. `CheckBalanceParams` data class
  4. `CapturePaymentParams` data class
2. Converts `ForageSDK` into a stateless class instead of a singleton object
3. Extends `ForageElement` interface to support `setForageContext(forageContext: ForageContext)`

Internally, this PR also introduces the `InternalForageElement` to keep common logic between `ForagePANEditText` and `ForagePINEditText` DRY.

## Test Plan
I updated the Sample App to use the v3 interface changes, tested locally, and everything worked fine. This is expected because `environment` is still coupled to `BuildConfig.FLAVOR`. Milestones 2 and 3 will break this dependency and is where more thorough testing will matter.

## How
Since we do not have Continuous Deployment, this PR (and the PRs below it) can be merged in when we are ready since the Sample App stays up to date with the changes introduced and it still follows the status-quo of relying on `BuildConfig.FLAVOR`

PR #103, opened 2023-09-21
Milestone 3 - Replace BuildConfig.FLAVOR with Stopgap FLAVOR

devinmorgan:
## What
This PR represents [Milestone 3](https://www.notion.so/joinforage/Plan-for-On-Time-Delivery-of-Android-3-0-Breaking-Changes-226d8ee6f8294d2694b1bb451791960b?pvs=4#66bf627d0ad34c50be03b8a002604f92) of the plan to deliver Android v3 changes to GoPuff by today

### Next Steps
- [x] Resolve all feedback
- [x] write up Migration Guide for GoPuff
- [ ] Release 🚀
- [ ] - [ ] update the Android changelog with these changes 

## Why
Reference the commit messages for a commit-by-commit breakdown of the code decisions that were made here.

## Test Plan
- ❌ I've added unit tests for this change. In the interest of expediency, we could get some fundamental confidence from the qa-tests passing and defer proper unit tests as tech debt
- ✅ The reviewer should manually test the changes in this PR. Yes, please do! This is the last of the Android v3 BREAKING changes so please test to your heart's content!

## Demo

https://github.com/teamforage/forage-android-sdk/assets/12377418/fc9c1c47-5b94-49e9-bf12-2b9f978593dd

## How
Once all the PRs in the stack are ✅, we can merge them into `main` one at a time, including this one.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
@@ -2,8 +2,8 @@ package com.joinforage.forage.android.collect
 
 import com.basistheory.android.service.BasisTheoryElements
 import com.basistheory.android.service.ProxyRequest
-import com.joinforage.forage.android.BuildConfig
 import com.joinforage.forage.android.core.Log
+import com.joinforage.forage.android.core.StopgapGlobalState
 import com.joinforage.forage.android.model.EncryptionKeys
 import com.joinforage.forage.android.model.PaymentMethod
 import com.joinforage.forage.android.network.ForageConstants
@@ -195,8 +195,11 @@ internal class BTPinCollector(
     }
 
     companion object {
-        private const val PROXY_ID = BuildConfig.BT_PROXY_ID
-        private const val API_KEY = BuildConfig.BT_API_KEY
+        // this code assumes that .setForageConfig() has been called
+        // on a Forage***EditText before PROXY_ID or API_KEY get
+        // referenced
+        private val PROXY_ID = StopgapGlobalState.envConfig.btProxyID
+        private val API_KEY = StopgapGlobalState.envConfig.btAPIKey
 
         private fun buildBt(): BasisTheoryElements {
             return BasisTheoryElements.builder()
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
@@ -1,8 +1,8 @@
 package com.joinforage.forage.android.collect
 
 import android.content.Context
-import com.joinforage.forage.android.BuildConfig
 import com.joinforage.forage.android.core.Log
+import com.joinforage.forage.android.core.StopgapGlobalState
 import com.joinforage.forage.android.model.EncryptionKeys
 import com.joinforage.forage.android.model.PaymentMethod
 import com.joinforage.forage.android.network.ForageConstants
@@ -204,8 +204,11 @@ internal class VGSPinCollector(
     }
 
     companion object {
-        private const val VAULT_ID = BuildConfig.VGS_VAULT_ID
-        private const val VGS_ENVIRONMENT = BuildConfig.VGS_VAULT_TYPE
+        // this code assumes that .setForageConfig() has been called
+        // on a Forage***EditText before PROXY_ID or API_KEY get
+        // referenced
+        private val VAULT_ID = StopgapGlobalState.envConfig.vgsVaultId
+        private val VGS_ENVIRONMENT = StopgapGlobalState.envConfig.vgsVaultType
 
         private fun buildVGSCollect(context: Context): VGSCollect {
             VGSCollectLogger.isEnabled = false
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
@@ -94,5 +94,9 @@ internal sealed class EnvConfig(
         fun fromForageConfig(context: ForageConfig?): EnvConfig {
             return fromSessionToken(context?.sessionToken)
         }
+
+        fun inProd(context: ForageContext?): Boolean {
+            return fromForageContext(context).FLAVOR == EnvOption.PROD
+        }
     }
 }
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
@@ -2,12 +2,12 @@ package com.joinforage.forage.android.core
 
 import com.joinforage.forage.android.ui.ForageConfig
 
-internal enum class EnvOption {
-    DEV,
-    STAGING,
-    SANDBOX,
-    CERT,
-    PROD
+internal enum class EnvOption(val value: String) {
+    DEV("dev"),
+    STAGING("staging"),
+    SANDBOX("sandbox"),
+    CERT("cert"),
+    PROD("prod")
 }
 
 internal sealed class EnvConfig(
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
@@ -91,12 +91,12 @@ internal sealed class EnvConfig(
             }
         }
 
-        fun fromForageConfig(context: ForageConfig?): EnvConfig {
-            return fromSessionToken(context?.sessionToken)
+        fun fromForageConfig(config: ForageConfig?): EnvConfig {
+            return fromSessionToken(config?.sessionToken)
         }
 
-        fun inProd(context: ForageContext?): Boolean {
-            return fromForageContext(context).FLAVOR == EnvOption.PROD
+        fun inProd(config: ForageConfig?): Boolean {
+            return fromForageConfig(config).FLAVOR == EnvOption.PROD
         }
     }
 }
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/StopgapGlobalState.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/StopgapGlobalState.kt
@@ -4,7 +4,6 @@ import com.joinforage.forage.android.ui.ForageConfig
 
 internal object StopgapGlobalState {
     var forageConfig: ForageConfig? = null
-
-    val FLAVOR: EnvOption
-        get() = EnvConfig.fromForageConfig(forageConfig).FLAVOR
+    val envConfig: EnvConfig
+        get() = EnvConfig.fromForageConfig(forageConfig)
 }
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt
@@ -9,8 +9,8 @@ import android.util.AttributeSet
 import android.util.TypedValue
 import com.google.android.material.textfield.TextInputEditText
 import com.google.android.material.textfield.TextInputLayout
-import com.joinforage.forage.android.BuildConfig
 import com.joinforage.forage.android.R
+import com.joinforage.forage.android.core.EnvConfig
 import com.joinforage.forage.android.core.Log
 import com.joinforage.forage.android.core.element.SimpleElementListener
 import com.joinforage.forage.android.core.element.StatefulElementListener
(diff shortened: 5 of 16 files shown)


devinmorgan:
* main
  * **PR #68** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/68" target="_blank">Graphite</a> 
    * **PR #93** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/93" target="_blank">Graphite</a> 
      * **PR #97** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/97" target="_blank">Graphite</a> 
        * **PR #98** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/98" target="_blank">Graphite</a> 
          * **PR #102** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/102" target="_blank">Graphite</a> 
            * **PR #103** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/103" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/103?utm_source=stack-comment).

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt
```diff
@@ -28,23 +28,21 @@ class ForagePANEditText @JvmOverloads constructor(
...
-        // allows whitelist of special Ebt PAN numbers
-        PanElementStateManager.NON_PROD_forEmptyInput()
-    }
-    private val formatTextWatcher: FormatPanTextWatcher
+
+    private var _SET_ONLY_manager: PanElementStateManager? = null
+    private val manager: PanElementStateManager
+        get() {
+            if (_SET_ONLY_manager == null) {
+                TODO("throw an error here since that means setForageContext was not called")
```
devinmorgan, line 36: Ideally I write a unit test suite that confirms that that `manager` and `_SET_ONLY_manager` work as expected similar to the the unit test suits we have in Forage JS. In the interest of expediency, I was going to gain confidence from qa-tests passing. But lmk what you guys think!

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
```diff
@@ -25,48 +25,71 @@ class ForagePINEditText @JvmOverloads constructor(
...
                     gravity = Gravity.CENTER
 
-                    val vaultType = LDManager.getVaultProvider(context.applicationContext as Application, logger)
-                    vault = if (vaultType == VaultConstants.BT_VAULT_TYPE) {
-                        BTVaultWrapper(context, attrs, defStyleAttr)
-                    } else {
-                        VGSVaultWrapper(context, attrs, defStyleAttr)
-                    }
+                    btVaultWrapper = BTVaultWrapper(context, attrs, defStyleAttr)
+                    vgsVaultWrapper = VGSVaultWrapper(context, attrs, defStyleAttr)
```
devinmorgan, line 50: probably should leave this as a commit (I definitely touch on it in the commit history) but we can initialize them both and can decide which one to add to the UI later!

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
```diff
@@ -25,48 +25,71 @@ class ForagePINEditText @JvmOverloads constructor(
...
+    private val vgsVaultWrapper: VGSVaultWrapper
+    private var _SET_ONLY_vault: VaultWrapper? = null
+    private val vault: VaultWrapper
+        get() {
+            if (_SET_ONLY_vault == null) {
+                TODO("throw an error here since that means setForageContext was not called")
+                throw Exception("need to call setForageContext first!")
+            }
+            return _SET_ONLY_vault!!
+        }
```
djoksimo, line 39: I left [a comment](https://github.com/teamforage/forage-android-sdk/pull/97#discussion_r1331955281) in your other PR about the dev experience and the risk of crashing in production if we throw in cases like this. 

Can you assure that this won't be an issue in production?

djoksimo: What is the `_SET_ONLY` member for? 
Instead of relying on `_SET_ONLY_vault` being set, can we make the `_config` member from the parent a protected member and check its value instead?

This cleanup work can be deferred to a separate ticket if this release is time-sensitive.

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
devinmorgan: Great question - I will definitely leave a comment to clarify. In the mean time, the relationship between _SET_ONLY_vault and vault is as follows:

The actual code in `ForagePINEditText` wants to use vault - that's
all it cares about. However, vault is the result of a decision
of whether we wish to use Basis Theory or VGS. The answer to this decision
depends on Launch Darkly. We can only get an answer from launch
darkly, once we know the environment we're in, which depends on
the forageConfig being set on this instance

Given the above, vault is an abstraction that allows the 
rest of the code in `ForagePANEditText` to work with a
certain (non-nullable) answer. It relies on the 
backing, write-only property `_SET_ONLY_vault`. `_SET_ONLY_vault`
will only be instantiated after ForageConfig was set on this instance
We enforce the abstraction that vault exists, by throwing a runtime
exception if vault, which is a computed property, was ever accessed
at a point `_SET_ONLY_vault` is null

devinmorgan: Regarding this:

> Instead of relying on _SET_ONLY_vault being set, can we make the _config member from the parent a protected member and check its value instead?

Let's circle back after the release to talk about this. I made a ticket here for it [here](https://linear.app/joinforage/issue/FX-588/android-consider-alternative-to-set-only-vault)

devinmorgan: Regarding this:

> the risk of crashing in production if we throw in cases like this.
>
> Can you assure that this won't be an issue in production?

the typical contract is that that everyone expects and is OK with is the contract
that the Android runtime enforces:

> You must initialize all of the state necessary for your class to work nominally
> within the `init { ... }` method.

As mentioned above:

> The actual code in ForagePINEditText wants to use vault - that's
> all it cares about. However, vault is the result of a decision
> of whether we wish to use Basis Theory or VGS. The answer to this decision
> depends on Launch Darkly. We can only get an answer from launch
> darkly, once we know the environment we're in, which depends on
> the forageConfig being set on this instance

This mean we need to make a new, more demanding contract than the typical one
developers expect:

> The Forage Android SDK guarantees that every ForageElement will work nominally
> if both `init { ... }` and `setForageConfig { ... }` get called prior before
> executing any other methods

This second contract, by definition of "nominal", guarantees that if merchants
call `setForageConfig { ... }` before they call anything else (recall that Android
runtime already calls `init { ... }`), no **unexpected** runtime exceptions will
be thrown.

Well that's all good in theory land but we care about in practice. The question
becomes how do we enforce the guarantee that if a merchant calls
`setForageConfig` before calling anything else, then no exceptions will be thrown?

The answer is basically enforcing conventions or Forage developers and a little
bit of type checking by the compiler:

Reducing Risk with Forage Developer Convention
`_SET_ONLY_vault` is a mutable variable and in theory any code in
`ForagePINEditText` could set it to `null` and that would break things. But, by
giving a variable such an ugly glaring name as _SET_ONLY_vault, the risk of such
an accident happening is significantly reduce (unlikely that your IDE will
suggest this variable if what you really want is to read `vault`) and the second
line of defense is that somebody reviewing a code change is likely to see an
ugly variable like `_SET_ONLY_vault` an raise a question.

Reducing Risk with Compile Time Guarantees
The `vault` variable that will be in high demand, is read-only so the
compiler enforces that it cannot be accidentally set to `null` (which would
totally lead to bugs otherwise).

We further reduce risk by not burdening downstream consumers of `vault` to have
to handle the possibility for `vault` being `null`. If every piece of code that
consumed `vault` needing to handle `null`, I agree that would create a risk that
an uncaught exception happens in `prod`. But, by throwing a runtime exception
if `_SET_ONLY_vault` is null in the `get()` definition of `vault`, the compiler
is acknowledges that there is no way any downstream code will accidentally see
`vault == null`.

You concern about the consequence of throwing unchecked exceptions at runtime is
legitimate. However, I feel our current strategy substantially reduces the risks
Let me know what you think and/or if you want to have a sync chat about this

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt
devinmorgan: I created a ticket for the unit tests [here](https://linear.app/joinforage/issue/FX-589/android-write-unit-tests-for-set-only-field-logic)

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
devinmorgan: created a comment in [b188107](https://github.com/teamforage/forage-android-sdk/pull/103/commits/b188107e9d92953befbecc503bbbffb52cb0a814)

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -195,8 +195,11 @@ internal class BTPinCollector(
     }
 
     companion object {
-        private const val PROXY_ID = BuildConfig.BT_PROXY_ID
-        private const val API_KEY = BuildConfig.BT_API_KEY
+        // this code assumes that .setForageContext() has been called
```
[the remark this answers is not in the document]
devinmorgan: Addressed in [eddbcce](https://github.com/teamforage/forage-android-sdk/pull/103/commits/eddbccef05a72c91910ee0fadd21c805c2970148). Thanks for flagging all of these!

forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
```diff
@@ -204,8 +204,11 @@ internal class VGSPinCollector(
     }
 
     companion object {
-        private const val VAULT_ID = BuildConfig.VGS_VAULT_ID
-        private const val VGS_ENVIRONMENT = BuildConfig.VGS_VAULT_TYPE
+        // this code assumes that .setForageContext() has been called
```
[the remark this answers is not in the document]
devinmorgan: Addressed in [eddbcce](https://github.com/teamforage/forage-android-sdk/pull/103/commits/eddbccef05a72c91910ee0fadd21c805c2970148). Thanks for flagging all of these!

forage-android/src/main/java/com/joinforage/forage/android/core/EnvConfig.kt
```diff
@@ -94,5 +94,9 @@ internal sealed class EnvConfig(
         fun fromForageContext(context: ForageContext?): EnvConfig {
             return fromSessionToken(context?.sessionToken)
         }
+
+        fun inProd(context: ForageContext?): Boolean {
+            return fromForageContext(context).FLAVOR == EnvOption.PROD
+        }
```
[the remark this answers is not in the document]
devinmorgan: Addressed in [eddbcce](https://github.com/teamforage/forage-android-sdk/pull/103/commits/eddbccef05a72c91910ee0fadd21c805c2970148). Thanks for flagging all of these!

forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePINEditText.kt
```diff
@@ -25,48 +25,71 @@ class ForagePINEditText @JvmOverloads constructor(
...
     }
 
+    override fun setForageContext(forageContext: ForageContext) {
+        // super is responsible for initializing the log and some
+        // global state so it must be called first
+        super.setForageContext(forageContext)
+
+        // Must initialize DD at the beginning of each render function. DD requires the context,
+        // so we need to wait until a context is present to run initialization code. However,
+        // we have logging all over the SDK that relies on the render happening first.
```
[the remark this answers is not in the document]
devinmorgan: OH I can't take credit for those comments. I think that was Danny I just copy/paste

forage-android/src/main/java/com/joinforage/forage/android/core/StopgapGlobalState.kt
```diff
@@ -3,8 +3,8 @@ package com.joinforage.forage.android.core
 import com.joinforage.forage.android.ui.ForageContext
 
 internal object StopgapGlobalState {
+    // TODO: change context -> forageConfig
     var context: ForageContext? = null
```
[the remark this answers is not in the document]
devinmorgan: should be addressed! see [here](https://github.com/teamforage/forage-android-sdk/pull/103/files#diff-1709252c8b7b6c42adf52258f2a517bebf9d536b1633b8de38467829cf0b8b0c)

PR #110, opened 2023-09-25
[FX-596] Make `setForageConfig` safe to call multiple times

devinmorgan:
## What
The linear ticket is [here](https://linear.app/joinforage/issue/FX-596/android-make-sure-that-setforageconfig-can-be-called-multiple-times). Also, see the commit messages for detailed explanations of changes

## Why
One of the design goals of introducing `setForageConfig` is to be able to handle `sessionToken` refresh or even update `merchantId` on the fly. This means `setForageConfig` needs to be safe to be called multiple times. Right now, it is not safe to call multiple times. It will actually crash the app:

image

## Test Plan
- ❌ I've added unit tests for this change.  In theory, there is logic in here that can be extracted out into a unit test. However, given the modifications this PR makes to `CatalogFragment.kt`, we gain some confidence that the logic in `AbstractForageElement` is working correctly. So, I decided to punt the unit tests for later
- ❌ The reviewer should manually test the changes in this PR. No need to test these changes manually because the CI pipeline combined with the modifications I made to `CatalogFragment.kt` will confirm this change works

## Demo
An image showcasing the catalog page working. This means it's not crashing and that `setForageConfig` can be called multiple times.
image

## How
This change can be released as is and in a patch level change.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/AbstractForageElement.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/AbstractForageElement.kt
@@ -12,13 +12,36 @@ abstract class AbstractForageElement(
 ) : LinearLayout(context, attrs, defStyleAttr), ForageElement {
 
     private var _forageConfig: ForageConfig? = null
+
+    // designated method that subclass can run all the
+    // side-effect things exactly once the first time
+    // setForageConfig is called
+    protected abstract fun initWithForageConfig()
+
     override fun setForageConfig(forageConfig: ForageConfig) {
+        // keep a record of whether this was the first time
+        // setForageConfig is getting called. we'll use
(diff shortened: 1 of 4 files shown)


devinmorgan:
* main
  * **PR #110** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/110" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/110?utm_source=stack-comment).

djoksimo (approved):

dleis612 (approved):

devinmorgan:
Going to merge this in without waiting for CI since CI passed before and I just updated a comment

PR #112, opened 2023-09-27
[FX-599] Don't hit VGS or BT when invalid PIN submitted

dleis612:
## What
If the PIN field is invalid when someone attempts to submit, don't hit our 3rd party APIs and instead just return that the PIN was invalid.

## Test Plan

- ❌ I've added unit tests for this change. -> I WILL BE ADDING A QA TEST FOR THIS
- ✅ The reviewer should manually test the changes in this PR.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
@@ -40,6 +40,27 @@ internal class VGSPinCollector(
         cardToken: String,
         encryptionKey: String
     ): ForageApiResponse<String> = suspendCoroutine { continuation ->
+        // If the PIN isn't valid (less than 4 numbers) then return a response here.
+        if (!pinForageEditText.getElementState().isComplete) {
+            logger.e(
+                "[VGS] User attempted to submit an invalid PIN",
+                attributes = mapOf(
+                    "merchant_ref" to merchantAccount,
+                    "payment_method_ref" to paymentMethodRef
+                )
+            )
+            continuation.resumeWith(
+                Result.success(
+                    ForageApiResponse.Failure(
+                        listOf(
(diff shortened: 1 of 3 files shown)


forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
```diff
@@ -173,6 +194,27 @@ internal class VGSPinCollector(
         cardToken: String,
         encryptionKey: String
     ): ForageApiResponse<String> = suspendCoroutine { continuation ->
+        // If the PIN isn't valid (less than 4 numbers) then return a response here.
+        if (!pinForageEditText.getElementState().isComplete) {
+            logger.e(
```
djoksimo, line 199:
```suggestion
            logger.w(
```

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/ForageConstants.kt
```diff
@@ -36,4 +37,10 @@ internal object ForageConstants {
     object VGS {
         const val PIN_FIELD_NAME = "pin"
     }
+
+    object ErrorResponseObjects {
+        val INVALID_PIN_ERROR = listOf(
```
djoksimo, line 42: folks can mistake this for INVALID_PIN, like `ebt_error_55` which could lead to misuse, can we use `INCOMPLETE_PIN_ERROR`?

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/ForageConstants.kt
dleis612: Why yes we can! :)

djoksimo (approved):

PR #113, opened 2023-09-27
[FX-597] Drop usage of Build Flavors in Forage Android

devinmorgan:
## What
The end goal is to drop the usage of build flavors from the Forage Android SDK to simplify the GoPuff declares their dependence on Forage Android SDK.

This PR is the first of the following two steps:
1. 👉 Drop build flavors in Forage Android SDK
2. Drop build flavors in the Sample App (and confirm that GoPuff can install the SDK without specifying a flavor)

See commit messages and comments for a detailed understanding of decisions made in this PR

## Why
Dropping flavors requires dropping our dependence on BuildConfig, which requires moving all BuildConfig values to the `EnvConfig` object. The last outstanding BuildConfig value was the version of the app which we use for logging.

## Test Plan

- ❌ I've added unit tests for this change.  No unit tests because this PR large touches build config things and constants
- ❌ The reviewer should manually test the changes in this PR.  No need to test manually as this is an intermediate story. The full change involves the entire stack.

## Demo
The demo is in the next PR!

## How
Because this PR and the next PR (#114 ) affect build pipeline things (dropping use on flavors), this PR and the next will break the CI and Lint checks. We should wait to merge this PR into `main` until we have merged the next PR and the subsequent PR (#115). Merging #113, #114, and #115 into `main` all at once is what we want to do to accomplish the changes while not breaking our GitHub checks in the process 👍

PR #114, opened 2023-09-27
Drop CreatePaymentViewModel dependence on BuildConfig

devinmorgan:
## What
The end goal is to drop the usage of build flavors from the Forage Android SDK to simplify the GoPuff declares their dependence on Forage Android SDK.

This PR is the second of the following two steps:
1. Drop build flavors in Forage Android SDK
2. 👉 Drop build flavors in the Sample App (and confirm that GoPuff can install the SDK without specifying a flavor)

## Why
GoPuff cares about us dropping flavors in `forage-android` and things will get a bit wonky if our `sample-app` still uses flavors (e.g. our CI workflows). Also, our `sample app` did not lean on flavors nearly as much as our `forage-android` leaned on flavors.

## Test Plan

- ❌ I've added unit tests for this change.  No unit tests for this change because it largely affected view logic or files that already lacked coverage or build files
- ✅ The reviewer should manually test the changes in this PR.  Not 100% necessary since the demo showcases it but it would be nice if somebody else confirmed on their machine that everything works locally without flavors

## Demo
This is a demo showcasing my local `sample-app` installing `forage-android` from a remote repository without any build flavors. Here's the breakdown:
1. it shows updating the sample app's `build.gradle` file to reference the local build of `forage-android`. This local build of android includes a `println()` 
2. It shows in the logs that the `println()` gets hit 
3. It then shows modifying the `build.gradle` to reference the remote repository (which is flavorless) and re-build the `sample-app
4. It shows that no `println()` gets hit in the logs (since it was not included in the remote build) and it showcases the entire flow working locally

https://github.com/teamforage/forage-android-sdk/assets/12377418/3eee75be-7779-4e66-a192-31035114d330

## How
Because this PR and the previous PR (#113 ) affect build pipeline things (dropping use on flavors), this PR and the previous are breaking the CI and Lint checks. We should wait to merge this PR and the previous into `main` until we have merged the next PR (#115). Merging #113, #114, and #115 into `main` all at once is what we want to do to accomplish the changes while not breaking our GitHub checks in the process 👍


devinmorgan:
* main
  * **PR #113** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/113" target="_blank">Graphite</a> 
    * **PR #114** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/114" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/114?utm_source=stack-comment).

sample-app/build.gradle
```diff
@@ -65,8 +46,12 @@ android {
 }
 
 dependencies {
+//    this line tests against the local forage-android
     implementation(project(":forage-android"))
 
+//     uncomment this line 👇 to test against remote repository forage-android
```
dleis612, line 52: Great addition!

sample-app/src/main/java/com/joinforage/android/example/ui/complete/flow/tokens/TokensViewModel.kt
```diff
@@ -2,14 +2,13 @@ package com.joinforage.android.example.ui.complete.flow.tokens
...
 import com.joinforage.android.example.ui.base.BaseViewModel
-import com.joinforage.android.example.ui.complete.flow.tokens.model.TokensUIDefaultState
 import dagger.hilt.android.lifecycle.HiltViewModel
 import javax.inject.Inject
 
 @HiltViewModel
 class TokensViewModel @Inject constructor() : BaseViewModel() {
-    val token = MutableLiveData(TokensUIDefaultState().bearer)
-    val merchantAccount = MutableLiveData(TokensUIDefaultState().merchantAccount)
+    val token = MutableLiveData("dev_1wtife770rjDwMUUbhe6jMZeEDVDtc")
```
dleis612, line 10: I think we don't want to put in a real token in the sample app

dleis612 (approved):

sample-app/src/main/java/com/joinforage/android/example/ui/complete/flow/tokens/TokensViewModel.kt
devinmorgan: great catch! I'll replace with a placeholder instead!

sample-app/build.gradle
```diff
@@ -65,8 +46,12 @@ android {
...
+//    implementation("com.joinforage:forage-android:3.1.0.1")
```
djoksimo, line 53: is it supposed to be `3.1.0` ?

devinmorgan: This was my attempt to make a version that GoPuff would not try to install while I was testing whether my changes to `CI.yaml` would work. I left it as a "placeholder" so to speak, so folks would replace it with any version they want to install from the repository. Clearly, the comment + fake version is not expressing my intent, so let me update the code to try and make the intent more clear. Thanks for flagging!

devinmorgan: Addressed in [f2b6450](https://github.com/teamforage/forage-android-sdk/pull/114/commits/f2b6450d4915b1dcb1c22b2d5f5257bb544cd433)

sample-app/src/main/java/com/joinforage/android/example/ui/complete/flow/tokens/TokensViewModel.kt
devinmorgan: Resolved in [03ebcd6](https://github.com/teamforage/forage-android-sdk/pull/114/commits/03ebcd676695ab408bd52f0f8baa89305a9da750)

PR #118, opened 2023-09-28
Create single source of truth for the Publish Version

devinmorgan:
## What
Danny spotted [a way](https://github.com/Basis-Theory/basistheory-android/blob/0a1435feb66c4732714ff9bb1b412636b63f7edf/lib/build.gradle#L16) for us to declare the `PUBLISH_VERSION` in exactly one place and have it exposed at run time via `BuildConfig`.

## Why
Have a single source of truth for the release version. Previously, we were updating two variables in the code with every release

## Test Plan

- ❌ I've added unit tests for this change.  No test because its a build related change
- ❌ The reviewer should manually test the changes in this PR.  No need since its a simple change and you can see that it works in the demo

## Demo
I tried to highlight the salient lines of code via the red breakpoints:
1. I changed the publish version to something new "3.3.0" which is different form our "3.20"
2. I'm exposing the PUBLISH_VERSION variable in the `defaultConfig` 
3. I've added a `println()` to `ForagePANEditText` that will print the value of `EnvConfig` if this technique
4. It prints so it works!!
image

## How
This PR can be merged to `main` once it's approved. It has no dependencies

### DIFF
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -5,6 +5,26 @@ plugins {
     id("org.jetbrains.kotlinx.kover") version "0.6.1"
 }
 
+ext {
+    PUBLISH_GROUP_ID = 'com.joinforage'
+    PUBLISH_VERSION = '3.3.0'
+    PUBLISH_ARTIFACT_ID = 'forage-android'
+    PUBLISH_DESCRIPTION = 'Forage Android SDK'
+    PUBLISH_URL = 'https://github.com/teamforage/forage-android-sdk'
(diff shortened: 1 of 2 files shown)


devinmorgan:
* main
  * **PR #118** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/118" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/118?utm_source=stack-comment).

dleis612 (approved):

PR #119, opened 2023-09-29
Danny/fx 543 make internal things internal in android sdk

dleis612:
## What
Based on the autogenerated docs, make all non-public things internal


djoksimo (approved):
Thank you @dleis612!

devinmorgan (approved):
This seems to good to me! That our Sample App passes in the `CI` check means the compiler didn't see anything wrong with what we're trying to expose publically and what we're trying to hide as `internal`. The only thing to brace for is that these were all public and so if GoPuff was using any of these, their builds will break. I can't see why they would use any of these but just pointing that out

djoksimo (approved):

PR #120, opened 2023-10-04
[FX-613] Remove throw statement and make error response consistent

dleis612:
## What
Remove the throw statement in the BT implementation that could potentially crash someone's app. Also, make the random response messages in the VGS implementation conform to the rest of the package.

## Testing
I still need to confirm that this change behaves as I want when Dendu makes a change to the BT PIN script on the backend.


dleis612:
@devinmorgan sooooooo

PR #121, opened 2023-10-04
[FX-580] Add trailing slash to all requests

dleis612:
## What
Add trailing slashes to all http requests that were missing them.

## Test Plan
I tested this locally by running through a checkout and looking at the datadog logs.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/EncryptionKeyService.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/EncryptionKeyService.kt
@@ -37,6 +37,7 @@ internal class EncryptionKeyService(
             .newBuilder()
             .addPathSegment(ForageConstants.PathSegment.ISO_SERVER)
             .addPathSegment(ForageConstants.PathSegment.ENCRYPTION_ALIAS)
+            .addPathSegment("")
             .build()
     }
 }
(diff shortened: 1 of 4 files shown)


forage-android/src/main/java/com/joinforage/forage/android/network/EncryptionKeyService.kt
```diff
@@ -37,6 +37,7 @@ internal class EncryptionKeyService(
             .newBuilder()
             .addPathSegment(ForageConstants.PathSegment.ISO_SERVER)
             .addPathSegment(ForageConstants.PathSegment.ENCRYPTION_ALIAS)
+            .addPathSegment("")
```
djoksimo, line 40: doing `addPathSegment("")` is not clear that we're adding a trailing slash. 

Would doing something like adding this extension work:

```kotlin
// in `OkHttpClientBuilder.kt` or new file: `HttpUrlExtensions.kt`
fun HttpUrl.Builder.addTrailingSlash(): HttpUrl.Builder {
    return this.addPathSegment("")
}

// in encryption key service

 return httpUrl
            .newBuilder()
            .addPathSegment(ForageConstants.PathSegment.ISO_SERVER)
            .addPathSegment(ForageConstants.PathSegment.ENCRYPTION_ALIAS)
            .addTrailingSlash()
            .build()
```

alternatively, can you add comments for each call to `addPathSegment("")`

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/EncryptionKeyService.kt
dleis612: This is a pretty cool optimization I didn't know was possible :)

dleis612: [ed1f578](https://github.com/teamforage/forage-android-sdk/pull/121/commits/ed1f578f596f6653b1b3cf5d53c311ad5c1d9990)

djoksimo (approved):

PR #122, opened 2023-10-17
Use datadog v2 in Log.kt

devinmorgan:
## What
Creates a custom Gradle Plugin that leverages the [Shadow Gradle Plugin](https://github.com/johnrengelman/shadow) to relocate an arbitrary AAR (e.g. datadog in our case) to a custom package namespace.

## Why
We want to continue using Datadog for our own analytics and telemetry. However, the current major version 1 of Datadog android uses a singleton pattern that ultimately leaks logs between us and merchants. The most straightforward next step was to install Datadog major version 2; however, this version has class name collisions with Datadog Major version 1 since a lot of code is shared between the v1 and v2.

We must move our datadog package into a new namespace to get around these class name collisions. This PR uses the Shadow Gradle Plugin to programmatically relocate the datadog v2 AAR to a different package namespace to avoid collisions. The upside of this approach is that there is a very small maintenance burden for continuing to use datadog. Bumping the package remains trivial, and this repo contains only our source code.

## Test Plan
I've tested this extensively myself. I've also included datadog v1 inside of the sample app so that the fact that all the CI checks are passing means that datadog v1 and our own datadog v2 can co-exist!

- ❌ I've added unit tests for this change.  No unit tests for now, though we could do that for the Gradle Plugin. I did not in the interest of time.
- ✅ The reviewer should manually test the changes in this PR.  Yes! Please test this locally if you are able. I want to ensure that nobody has any Java version issues since I upgraded some core packages.

## Demo

A full walkthrough of the checkout flow with a side-by-side of the logs in Datadog. So far, everything seems to work!

https://github.com/teamforage/forage-android-sdk/assets/12377418/8d86bccf-1145-4706-9906-cd812fa2a92d

## How
This PR can be merged in as soon as it get's approvals

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,9 +6,9 @@ buildscript {
 }
 
 plugins {
-    id 'com.android.application' version '7.3.1' apply false
-    id 'com.android.library' version '7.3.1' apply false
-    id 'org.jetbrains.kotlin.android' version '1.7.21' apply false
+    id 'com.android.application' version '8.1.0' apply false
+    id 'com.android.library' version '8.1.0' apply false
+    id 'org.jetbrains.kotlin.android' version '1.8.10' apply false
     id "com.diffplug.spotless" version "6.11.0"
     id 'androidx.navigation.safeargs' version '2.5.3' apply false
     id 'com.google.dagger.hilt.android' version '2.44' apply false
@@ -24,6 +24,15 @@ spotless {
     kotlinGradle {
         target '*.gradle.kts'
         ktlint()
+    }
+}
+
+
+allprojects {
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8.toString()
+        }
     }
 }
 
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -1,7 +1,7 @@
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
-    id("com.google.devtools.ksp").version("1.7.21-1.0.8")
+    id("com.google.devtools.ksp").version("1.8.10-1.0.9")
     id("org.jetbrains.kotlinx.kover") version "0.6.1"
 }
 
@@ -55,6 +55,7 @@ android {
 
     buildFeatures {
         dataBinding true
+        buildConfig true
     }
 
     testOptions {
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -1,9 +1,12 @@
+import com.joinforage.ShadowAarDependenciesPlugin
+
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
     id("com.google.devtools.ksp").version("1.8.10-1.0.9")
     id("org.jetbrains.kotlinx.kover") version "0.6.1"
 }
+apply plugin: ShadowAarDependenciesPlugin
 
 ext {
     PUBLISH_GROUP_ID = 'com.joinforage'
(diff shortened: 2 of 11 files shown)


devinmorgan:
* main
  * **PR #122** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/122" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/122?utm_source=stack-comment).

build.gradle
```diff
@@ -27,4 +27,13 @@ spotless {
     }
 }
 
+
+allprojects {
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
+        kotlinOptions {
+            jvmTarget = JavaVersion.VERSION_1_8.toString()
```
dleis612, line 34: Is it a big deal that we are upgrading our JVM?

dleis612 (approved):
This is awesome! Can I book some time with you just to run through how this works out of curiosity?

build.gradle
```diff
@@ -6,9 +6,9 @@ buildscript {
 }
 
 plugins {
-    id 'com.android.application' version '7.3.1' apply false
-    id 'com.android.library' version '7.3.1' apply false
-    id 'org.jetbrains.kotlin.android' version '1.7.21' apply false
+    id 'com.android.application' version '8.1.0' apply false
+    id 'com.android.library' version '8.1.0' apply false
```
devinmorgan, line 10: Bumping the Android Gradle Plugin here to major version 8 [requires JDK 17](https://developer.android.com/build/releases/gradle-plugin#:~:text=Android%20Gradle%20plugin%208.1.0%20is%20a%20major%20release%20that%20includes%20a%20variety%20of%20new%20features%20and%20improvements.)

djoksimo (changes requested):

forage-android/build.gradle
```diff
@@ -68,6 +72,45 @@ android {
...
+     * Gradle will also take care of recurisvely fetching any sub-dependencies
+     * as well.
+     *
+     * Because of the intensive manipulations we are doing to the datadog AAR
+     * artificat, we have yet to figure out an elegant way for Gradle to continue
+     * handling the process for resolving the dependencies and sub-dependencies
+     * of datadog-android-sdk-logs.
+     *
+     * Until we have a more elquent answer, we are opting to manually install
+     * all the sub-denpendencies that datadog-android-sdk-logs and its peer
```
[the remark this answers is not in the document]
devinmorgan: ahh thank you! Yeah I'll give the comments another look-over for typos etc. Appreciate these callouts!

build.gradle
devinmorgan: Danny! Thank you for your patience - I was trying to investigate this to get you an intelligent answer. 

So to clarify I see that this comment was left near this code block
```groovy
allprojects {
    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
        kotlinOptions {
            jvmTarget = JavaVersion.VERSION_1_8.toString()
```
but I know elsewhere I changed the CI to use the JDK 17, an upgrade from JDK 11.

**Short answer: yes, all of this should be fine**

**Longer answer below 👇**
For the sake of this discussion, there are three concepts that matter:
1. What version of Java Gradle is going to use to run the build pipeline
2. What version of the Java API and libraries we want to use in our source code
3. What version of Java byte code we want to compile our source code

Starting with #3, Android basically only supports Java 8 and lower, so we need to (and have been) set `targetCompatibility = Java 8`. 
Moving to #2, the version of the Java API that we want to use in our source code can be as low as we want but no higher than the version of Java bundled into the JDK that we are using (previously JDK 11, which has Java 11 but now upgraded to JDK 17, which has Java 17). The JDK will take care of compiling our Java 17 into Java 8 byte code.
Moving to #1, this is the version of the JDK that we have installed in our CI action or on our local machine. The reason that we need to use JDK 17 is actually because the Android Gradle Plugin version 8 specifying it needs JDK 17

So we end up with
* Android Gradle Plugin tells us to use JDK 17 so we install it in our CI and our local machines
* leads to Gradle using JDK 17 since that's what's available in CI / local machines
* Android devices run on Java 8 so we tell `targetCompatibility` to be Java 8
* Source code we limit to Java 8 mostly just because though it could go as high as Java 17 since we have JDK 17 installed

Lastly this block here is just to remind some tasks in Gradle that we are trying to target Java 8 byte code that's all
```groovy
allprojects {
    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
        kotlinOptions {
            jvmTarget = JavaVersion.VERSION_1_8.toString()
```

djoksimo (commented):
Should we update this part of the README to say JDK 17?

> Technically, our CI/CD builds Forage Android SDK using [JDK 11](https://github.com/teamforage/forage-android-sdk/blob/f2ff61ed27847b28d631f975c9a79e52f9258802/.github/workflows/CI.yaml#L21).

devinmorgan:
> Should we update this part of the README to say JDK 17?
> 
> > Technically, our CI/CD builds Forage Android SDK using [JDK 11](https://github.com/teamforage/forage-android-sdk/blob/f2ff61ed27847b28d631f975c9a79e52f9258802/.github/workflows/CI.yaml#L21).

Yes we definitely should! That was on my list and forgot about it so thanks for surfacing. Will make that update shortly!

djoksimo (approved):

PR #126, opened 2023-11-06
[FX-677] Remove ForageConstants dependence on StopgapGlobalState

devinmorgan:
## What
This PR is associated with [removing `ForageConstant.kt`'s dependence on `StopgapGlobalState`](https://www.notion.so/joinforage/Pre-GoPuff-Launch-Cleanup-Proposal-aa5c5caf4a6747fc930ad511e83b79b6?pvs=4#357ef83068174aabba9256e14e36198b), part of the remaining cleanup work associated with Android 3.0

`ForageConstant.kt`'s depended on `StopgapGlobalState` because of `BASE_URL`. However, `BASE_URL` is only directly consumed by `ForageSDK`, which means we make moving the related logic directly into `ForageSDK` by parsing the `sessionToken` to extract the `baseUrl`

## Test Plan
- ✅ and ❌. I added unit tests to ensure surface if any of the `EncConfig` `baseUrl`s ever become malformed because we are unable to get the compiler's guarantee of non-null values or that it won't throw runtime errors. We had a thoughtful conversation about this [below](https://github.com/teamforage/forage-android-sdk/pull/126#discussion_r1383861013)
- ❌ It's not necessary to manually test these changes as the CI tests will flag if there's an issue with these changes.

## How
This can be rolled out once approved!

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/PaymentMethodService.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/PaymentMethodService.kt
@@ -5,12 +5,13 @@ import com.joinforage.forage.android.core.telemetry.Log
 import com.joinforage.forage.android.network.model.ForageApiResponse
 import com.joinforage.forage.android.network.model.ForageError
 import okhttp3.HttpUrl
+import okhttp3.HttpUrl.Companion.toHttpUrlOrNull
 import okhttp3.OkHttpClient
 import okhttp3.Request
 import java.io.IOException
 
 internal class PaymentMethodService(
-    private val httpUrl: HttpUrl,
+    private val httpUrl: String,
     okHttpClient: OkHttpClient,
     private val logger: Log
 ) : NetworkService(okHttpClient, logger) {
@@ -43,13 +44,11 @@ internal class PaymentMethodService(
         return convertCallbackToCoroutine(request)
     }
 
-    private fun getPaymentMethodUrl(paymentMethodRef: String): HttpUrl {
-        return httpUrl
-            .newBuilder()
-            .addPathSegment(ForageConstants.PathSegment.API)
-            .addPathSegment(ForageConstants.PathSegment.PAYMENT_METHODS)
-            .addPathSegment(paymentMethodRef)
-            .addTrailingSlash()
-            .build()
-    }
+    private fun getPaymentMethodUrl(paymentMethodRef: String): HttpUrl = httpUrl.toHttpUrlOrNull()!!
+        .newBuilder()
+        .addPathSegment(ForageConstants.PathSegment.API)
+        .addPathSegment(ForageConstants.PathSegment.PAYMENT_METHODS)
+        .addPathSegment(paymentMethodRef)
+        .addTrailingSlash()
+        .build()
 }
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
@@ -1,11 +1,11 @@
 package com.joinforage.forage.android
 
+import com.joinforage.forage.android.core.EnvConfig
 import com.joinforage.forage.android.core.telemetry.CustomerPerceivedResponseMonitor
 import com.joinforage.forage.android.core.telemetry.EventOutcome
 import com.joinforage.forage.android.core.telemetry.Log
 import com.joinforage.forage.android.core.telemetry.UserAction
 import com.joinforage.forage.android.network.EncryptionKeyService
-import com.joinforage.forage.android.network.ForageConstants
 import com.joinforage.forage.android.network.MessageStatusService
 import com.joinforage.forage.android.network.OkHttpClientBuilder
 import com.joinforage.forage.android.network.PaymentMethodService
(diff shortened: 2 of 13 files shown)


devinmorgan:
* main
  * **PR #126** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/126" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/126?utm_source=stack-comment).

forage-android/src/main/java/com/joinforage/forage/android/network/PaymentMethodService.kt
```diff
@@ -43,13 +44,11 @@ internal class PaymentMethodService(
...
-    private fun getPaymentMethodUrl(paymentMethodRef: String): HttpUrl {
-        return httpUrl
-            .newBuilder()
-            .addPathSegment(ForageConstants.PathSegment.API)
-            .addPathSegment(ForageConstants.PathSegment.PAYMENT_METHODS)
-            .addPathSegment(paymentMethodRef)
-            .addTrailingSlash()
-            .build()
-    }
+    private fun getPaymentMethodUrl(paymentMethodRef: String): HttpUrl = httpUrl.toHttpUrlOrNull()!!
```
djoksimo, line 47: can we refactor this to avoid force unwrapping here?

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/PaymentMethodService.kt
devinmorgan: Hey Danilo! I agree with you that we want to squash the risk of a non-null assertion like leading to an accidental runtime error. Here's the situation from what I can see:

The value that get's passed for `httpUrl` is the [`baseUrl` as noted here](https://github.com/teamforage/forage-android-sdk/blob/1048904bea39af315b1e2c566846f5e8c5304be2/sample-app/src/main/java/com/joinforage/android/example/network/model/EnvConfig.kt#L16-L39). To your point, the CI-tests will only surface that if the `staging` URL is broken. Further, local development would likely only surface if `sandbox` or `dev`'s URLs are broken. That means, technically, we don't currently have an automated way of ensuring that the `prod` URL is correct / well-formed.

I'd be happy to write a test [against `EnvConfig`](https://github.com/teamforage/forage-android-sdk/blob/1048904bea39af315b1e2c566846f5e8c5304be2/sample-app/src/main/java/com/joinforage/android/example/network/model/EnvConfig.kt#L11) that ensures that the values of `baseUrl` are correct for each environment. This would give us an automated way of knowing if somebody accidentally broke a URL. However, I don't think outside of an automated test we'd be able to get the compiler's guarantee that a URL will never be null. Allow me to explain:

Here is, for example, an alternative way these get-URLs could be written:
```kotlin
private fun getTokenizeCardUrl(): HttpUrl = Request.Builder()
        .url(httpUrl)
        .build()
        .url
        .newBuilder()
        .addPathSegment(ForageConstants.PathSegment.API)
        .addPathSegment(ForageConstants.PathSegment.PAYMENT_METHODS)
        .addTrailingSlash()
        .build()
```
While the above technically makes the compiler happy that we're working with a non-null type, the [function specs](https://github.com/square/okhttp/blob/c73f9842ad72e4169f2c2f149b5bdee83fb73a76/okhttp/src/jvmMain/kotlin/okhttp3/Request.kt#L191-L196) still say that it will throw an `IllegalArgumentException` if the passed URL is not valid.

Basically the situation seems be that if the URL ever breaks, we'll get a possibly-null variable (with the current implementation) or we can use a function that says it will throw at run time.

My take on take on the best path forward is to add automated tests that ensure we don't break any URLs but let me know what you think!

djoksimo: Thank you for the thoughtful insights!

> My take on take on the best path forward is to add automated tests that ensure we don't break any URLs but let me know what you think!

I'm on board with this!

devinmorgan: Added automated tests in [d505d60](https://github.com/teamforage/forage-android-sdk/pull/126/commits/d505d6079c057423067015ab3e1d2bc18be453df)!

djoksimo: Thank you!

djoksimo (commented):
Thank you for adding the tests and for the overall code improvements in this PR stack!

Do you have thoughts/concerns about the coverage diff CI checks across this stack reporting a bit under 80% (76, 77, ...) – if we're aiming to increase the overall code coverage?

devinmorgan:
> Do you have thoughts/concerns about the coverage diff CI checks across this stack reporting a bit under 80% (76, 77, ...) – if we're aiming to increase the overall code coverage?

You're getting excited here talking about possibly needing to add more tests haha 🤤

Just to make sure that I am on the same page as you, here is my understanding of the CI diff:

image

All of these files have 100% coverage:
- `EncryptionKeyService.kt`
- `MessageStatusService.kt`
- `PaymentMethodService.kt`
- `PaymentService.kt`
- `TokenizeCardService.kt`

and `ForageSDK` has 0% coverage which averages 77% coverage for the diff.

With that as my understanding, my take is that yes it would be awesome to have 80% or more coverage on `ForageSDK`! I think this may be best done in a follow on PR though. The reason I'd say a follow on PR may be more appropriate is because addin any amount of meaningful coverage to `ForageSDK` would involve some non-trivial refactors so that `ForageSDK` and it's methods can be tested at all - right now `ForageSDK` isn't tested at all.

Let me know what you think!

djoksimo (approved):

dleis612 (approved):

PR #127, opened 2023-11-07
[FX-678] Remove Forage Log's dependence on StopgapGlobalState

devinmorgan:
## What
This PR is associated with [removing `Log.kt`'s dependence on `StopgapGlobalState`](https://www.notion.so/joinforage/Pre-GoPuff-Launch-Cleanup-Proposal-aa5c5caf4a6747fc930ad511e83b79b6?pvs=4#13dcc917df71402190f0b96fd8a4ac2a), part of the remaining cleanup work associated with Android 3.0

`Log.kt`'s depended on `StopgapGlobalState` because of the Datadog client key and we were able to remove that dependence by having `Log.initializeDD(...)` accept a `ForageConfig` as a param

## Test Plan
- ❌.  No novel unit tests were add but I updated the unit tests in accordance with the changes made
- ❌ It's not necessary to manually test these changes as the CI tests will flag if there's an issue with these changes.

## How
This can be rolled out once approved!

PR #128, opened 2023-11-07
[FX-681] Remove LDManager's dependence on StopgapGlobalState

devinmorgan:
## What
This PR is associated with [removing `LDManager.kt`'s dependence on `StopgapGlobalState`](https://www.notion.so/joinforage/Pre-GoPuff-Launch-Cleanup-Proposal-aa5c5caf4a6747fc930ad511e83b79b6?pvs=4#31c155cd4bdb40d78f8e78295c7b2bc5), part of the remaining cleanup work associated with Android 3.0

`LDManage.kt` depended on `StopgapGlobalState` for the Launch Darkly client key. We were able to remove that dependence by having `LDMangager.initialize(...)` accept an `LDConfig`, which will have the LD client key already set.

## Test Plan
- ❌ While the code coverage will not change with this PR, I did introduce two additional unit tests to explicitly test the helpder function responsible for mapping the primary vault traffic flag to a Vault Provider
- ❌ It's not necessary to manually test these changes as the CI tests will flag if there's an issue with these changes.

## How
This can be rolled out once approved!

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
@@ -1,7 +1,6 @@
 package com.joinforage.forage.android
 
 import android.app.Application
-import com.joinforage.forage.android.core.StopgapGlobalState
 import com.joinforage.forage.android.core.telemetry.Log
 import com.launchdarkly.sdk.ContextKind
 import com.launchdarkly.sdk.LDContext
@@ -32,40 +31,39 @@ internal object LDContextKind {
     const val SERVICE = "service"
 }
 
-internal object LDManager {
-    private val LD_MOBILE_KEY = StopgapGlobalState.envConfig.ldMobileKey
-    private var internalVaultType: VaultType? = null
+internal val ALWAYS_BT = 100.0
+internal val ALWAYS_VGS = 0.0
 
+internal fun computeVaultType(trafficPrimaryPercentFlag: Double): VaultType {
+    val randomNum = Math.random() * 100
+    return if (randomNum < trafficPrimaryPercentFlag) VaultType.BT_VAULT_TYPE else VaultType.VGS_VAULT_TYPE
+}
+
+internal object LDManager {
     private var internalLogger: Log = Log.getSilentInstance()
     private var client: LDClient? = null
 
-    internal var vaultType: VaultType?
-        get() = internalVaultType
+    // default to 100% VGS usage in case LD flag retrieval fails
+    private var primaryTrafficPercent = ALWAYS_VGS
 
-        // The setter is only exposed for testing purposes. Otherwise, it is entirely internal and
-        // shouldn't be used directly.
-        set(value) {
-            if (value == null) {
-                internalLogger.w("[LaunchDarkly] vaultType is being reset to null. This should only happen while unit testing!")
-                internalVaultType = null
-            } else if (internalVaultType == null) {
-                internalVaultType = value
-            } else {
-                throw Error("vaultType can only be set once!")
-            }
-        }
+    // we only ever want to fetch primary traffic percent flag once
+    // so that all parts of the code are guaranteed to use the same
+    // value
+    private var hasFetchedPrimaryTrafficPercentFlag = false
 
-    internal fun initialize(app: Application, logger: Log, dataSource: TestData? = null) {
-        val ldConfig = if (dataSource != null) {
-            LDConfig.Builder()
-                .mobileKey(LD_MOBILE_KEY)
-                .dataSource(dataSource)
-                .build()
-        } else {
-            LDConfig.Builder()
-                .mobileKey(LD_MOBILE_KEY)
-                .build()
-        }
+    fun createLdConfig(ldMobileKey: String): LDConfig {
+        return LDConfig.Builder()
+            .mobileKey(ldMobileKey)
+            .build()
+    }
+    fun TEST_createLdConfig(ldMobileKey: String, dataSource: TestData): LDConfig {
+        return LDConfig.Builder()
+            .mobileKey(ldMobileKey)
+            .dataSource(dataSource)
+            .build()
+    }
+
+    internal fun initialize(app: Application, logger: Log, ldConfig: LDConfig) {
         val contextKind = ContextKind.of(LDContextKind.SERVICE)
         val context = LDContext.create(contextKind, LDContexts.ANDROID_CONTEXT)
         client = LDClient.init(app, ldConfig, context, 0)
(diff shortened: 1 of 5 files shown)


devinmorgan:
* main
  * **PR #126** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/126" target="_blank">Graphite</a> 
    * **PR #127** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/127" target="_blank">Graphite</a> 
      * **PR #128** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/128" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/128?utm_source=stack-comment).

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -32,68 +31,78 @@ internal object LDContextKind {
...
         val context = LDContext.create(contextKind, LDContexts.ANDROID_CONTEXT)
         client = LDClient.init(app, ldConfig, context, 0)
 
         internalLogger = logger
     }
 
-    // vaultType is instantiated lazily and is a singleton. Once we set the vault type once, we don't
-    // want to overwrite it! We must take in the application as a parameter, which means that a
-    // ForagePINEditText must be rendered before any of the ForageSDKApi functions are called.
+    internal fun TEST_clearPrimaryTrafficPercentCache() {
```
djoksimo, line 74: The original implementation was intentionally written to avoid having test-use-only methods – which I saw as an antipattern/smell.

Can you elaborate on the benefits of this `TEST_` approach over the previous implementation?

In the iOS SDK, we use a `useCache: Boolean = true` to test against the cache-less code path

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -32,68 +31,78 @@ internal object LDContextKind {
     const val SERVICE = "service"
 }
 
-internal object LDManager {
-    private val LD_MOBILE_KEY = StopgapGlobalState.envConfig.ldMobileKey
-    private var internalVaultType: VaultType? = null
+internal val ALWAYS_BT = 100.0
+internal val ALWAYS_VGS = 0.0
```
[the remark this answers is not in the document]
devinmorgan: good suggestion!

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -32,68 +31,78 @@ internal object LDContextKind {
...
-                internalLogger.w("[LaunchDarkly] vaultType is being reset to null. This should only happen while unit testing!")
-                internalVaultType = null
-            } else if (internalVaultType == null) {
-                internalVaultType = value
-            } else {
-                throw Error("vaultType can only be set once!")
-            }
-        }
+    // default to 100% VGS usage in case LD flag retrieval fails
+    private var primaryTrafficPercent = ALWAYS_VGS
```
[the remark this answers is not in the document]
devinmorgan: Good question! I was inspired by the point you surfaced in [the Minimize SDK Failures design doc](https://docs.google.com/document/d/1lO5uFpYwkE-AUYKPWBrWtX90mNVYiCeqkGV1CySMK30/edit#bookmark=id.6nweccylo1j4) to refactor:

`internal var vaultType: VaultType?` ➡️ `private var primaryTrafficPercent = ALWAYS_VGS`

To expand on this, `vaultType` was able to be null which required us to do type-coercion [here](https://github.com/teamforage/forage-android-sdk/blob/1048904bea39af315b1e2c566846f5e8c5304be2/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt#L96). Here's a screenshot showing the IDE complaint

image

In other words, I wanted to eliminate the ability to do `LDManager.vaultType = null` to prevent the possibily of `getVaultProvider` ever accidentally return `null`. Using a non-null `primaryTrafficPercent` variable has no risk `getVaultProvider` returning `null` because `primaryTrafficPercent` is always guaranteed to be a Float.

That was my rationale but let me know your thoughts!

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
devinmorgan: Great point! I mentioned this on standup today but your question hear motivated me to look a bit deeper into why we were even caching `getVaultProvider` in the first place. It turns out we no longer need to do this so I updated the PR to remove all test-specific functionality. Not sure if the same came be done in iOS but glad you brought this up - `LDManager` should be much simpler now!

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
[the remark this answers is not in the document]
devinmorgan: addressed!

djoksimo (approved):

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -32,84 +30,60 @@ internal object LDContextKind {
...
+internal fun computeVaultType(trafficPrimaryPercentFlag: Double): VaultType {
+    val randomNum = Math.random() * 100
+    return if (randomNum < trafficPrimaryPercentFlag) VaultType.BT_VAULT_TYPE else VaultType.VGS_VAULT_TYPE
+}
 
-    private var internalLogger: Log = Log.getSilentInstance()
+internal object LDManager {
+    // as much as I would LOVE to not have this be a shared state on a
+    // singleton object, Launch Darkly literally tells us to make it
+    // a singleton :/
```
dleis612, line 44: :(

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -32,84 +30,60 @@ internal object LDContextKind {
...
+            LDFlags.VAULT_PRIMARY_TRAFFIC_PERCENTAGE_FLAG,
+            ALWAYS_VGS_PERCENT
+        ) ?: ALWAYS_VGS_PERCENT
+        logger.i("[LaunchDarkly] Vault percent of $vaultPercent return from LD")
+
+        // convert the flag percent into an answer to which vault provider to use
+        val vaultType = computeVaultType(vaultPercent)
+        logger.i("[LaunchDarkly] Vault type set to $vaultType")
+
+        // return vault provider derived from the novel or cached flag value
```
dleis612, line 66: We don't cache that value anymore right?

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -32,84 +30,60 @@ internal object LDContextKind {
...
+            defaultVal
+        ) ?: defaultVal
 
         val intervals = jsonIntervals.get("intervals")
 
         // Converting the LDArray into a LongArray
         val pollingList = LongArray(intervals.size()) { intervals.get(it).longValue() }
 
-        internalLogger.i("[LaunchDarkly] polling intervals $pollingList")
+        logger.i("[LaunchDarkly] polling intervals $pollingList")
```
dleis612, line 86: [This](https://app.datadoghq.com/logs?query=service%3Aandroid-sdk%20%22%5BLaunchDarkly%5D%20polling%20intervals%22%20&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=paused&stream_sort=desc&viz=stream&from_ts=1699602086056&to_ts=1699623600000&live=false) is what the log of polling intervals currently looks like :( looks like we need some kind of `toString` method?

forage-android/src/test/java/com/joinforage/forage/android/network/data/LaunchDarklyTest.kt
```diff
@@ -1,84 +1,77 @@
...
 
-@OptIn(ExperimentalCoroutinesApi::class)
+val LD_MOBILE_KEY = "some key"
+
 @RunWith(RobolectricTestRunner::class)
 class LaunchDarklyTest() {
-    private val alwaysBT = 100.0
-    private val alwaysVGS = 0.0
 
+    // NOTE: It is essential that the TestData source live in a
```
dleis612, line 25: Just a note on this, I had issues with these tests running with shared state when I was originally creating them. I don't have a full understanding of Kotlin, but I believe Kotlin companion objects are static singletons "whose properties and functions are tied to a class but not to the instance of that class". I guess that's why this causes the tests to function correctly? The more ya know 🤷‍♂️

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
devinmorgan: roger that! If it's cool with you, I made a follow on ticket for this [here](https://linear.app/joinforage/issue/FX-718/create-a-tostring-method-for-dollarpollinglist) just to keep that fix isolated

forage-android/src/test/java/com/joinforage/forage/android/network/data/LaunchDarklyTest.kt
devinmorgan: That would make sense but also i have NO IDEA lmao. Figure I at least document the bizarre behavior in a comment for anybody else modifying the tests

forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
devinmorgan: fixed in [a76ec00](https://github.com/teamforage/forage-android-sdk/pull/128/commits/a76ec003a8aaf3e2d8f8e10856294dc92e19262a)

dleis612 (approved):

PR #132, opened 2023-12-01
[FX-772] Add derivedCardInfo

dleis612:
## Test Plan


forage-android/src/main/java/com/joinforage/forage/android/ui/DerivedCardInfo.kt
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * The US state that issued the EBT card, derived from the Issuer Identification Number (IIN),
+     * also known as BIN (Bank Identification Number). The IIN is the first 6 digits of the PAN.
+     */
+    var usState: USState?
+}
+
+class DerivedCardInfo(usState: USState? = null) : DerivedCardInfoInterface {
+    override var usState: USState? = usState
+}
```
devinmorgan, line 19: **Possible opportunity to simply things using Kotlin conventions**

Defining an interface and then having a class that implements that interface is almost always the way to go. In Kotlin, my understanding is that it's convention for simple data like `DerivedCardInfo` to use `data class` instead of an `interface` + `class` (which is the way to go in almost all other languages).

I've found that if I'm defining `interface SomeEntityInterface {...}` and I really only ever anticipate 1 class ever implementing this interface called `class SomeEntity : SomeEntityInterface {...}`, that using a `data class SomeEntity {...}` may be the way to go instead.

As a reference, we faced a similar tradeoff in the [`ForageElement` file](https://github.com/teamforage/forage-android-sdk/blob/f430751747f0e8887ab6ca218c8db92a7552d1ef/forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt). Here, `ForageConfig` a simple data object and at the time we started doing `interface ForageConfigInterface {...}` + `class ForageConfig : ForageConfigInterface {...}` but ultimately switched to `data class ForageConfig`. `ForageElement` however is much more complex and we already know that we want `ForagePANEditText` and `ForagePINEditText` to adhere to it so the `interface` +`class` pattern felt much much more appropriate there.

```suggestion
data class DerivedCardInfo(
    /**
     * The US state that issued the EBT card, derived from the Issuer Identification Number (IIN),
     * also known as BIN (Bank Identification Number). The IIN is the first 6 digits of the PAN.
     */
    val usState: USState? = null
)
```

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/PanElementStateManager.kt
```diff
@@ -90,6 +92,12 @@ internal class PanElementStateManager(state: ElementState, private val validator
             .firstOrNull { it != null }
     }
 
+    private fun inferCardInfo(cardNumber: String): DerivedCardInfo {
+        return DerivedCardInfo(
```
devinmorgan, line 96: This is great! Love that we're returning a new instance of `DerivedCardInfo` each time instead of mutating an existing instance. Nice!!

forage-android/src/main/java/com/joinforage/forage/android/ui/DerivedCardInfo.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Represents card information derived from the user's current
+ * Primary Account Number (PAN) input value.
+ */
+interface DerivedCardInfoInterface {
+    /**
+     * The US state that issued the EBT card, derived from the Issuer Identification Number (IIN),
+     * also known as BIN (Bank Identification Number). The IIN is the first 6 digits of the PAN.
+     */
+    var usState: USState?
```
devinmorgan, line 14: Any particular reason we wanted this to be `var` (mutable)? From what I can see, you've been using `DerivedCardInfo.usState` in a strictly read-only / immutable way so far (which is awesome!). Could be a nice little win to make it `val` instead.

devinmorgan (commented):
Danny! This PR is sick. Nice turnaround. Left a few comments for your consideration so let me know what you think. Thanks for doing this!

PR #134, opened 2023-12-01
Update sample app to show usState of PAN

devinmorgan:
## What
Update the sample app to leverage the newly exposed `state.details.derivedCardInfo.usState` field.

## Why
Keep the sample app in sync with visual functionality that our customers can use.

## Test Plan

- ❌ Not unit tests because this is the Sample App, which does not support unit tests at present
- ✅ I'd recommend playing around with it just to see but the demo should suffice.

## Demo
Sample app has been updated to display the `usState`

https://github.com/teamforage/forage-android-sdk/assets/12377418/1c15b6bc-8575-48c3-9f29-3aaa6c187f51

## How
This PR needs to be merged into #133. It is not standalone.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
@@ -49,7 +49,7 @@ internal object LDManager {
     internal fun initialize(app: Application, ldConfig: LDConfig) {
         val contextKind = ContextKind.of(LDContextKind.SERVICE)
         val context = LDContext.create(contextKind, LDContexts.ANDROID_CONTEXT)
-        client = LDClient.init(app, ldConfig, context, 0)
+        client = LDClient.init(app, ldConfig, context, 1)
     }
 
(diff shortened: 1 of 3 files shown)


forage-android/src/main/java/com/joinforage/forage/android/LDManager.kt
```diff
@@ -49,7 +49,7 @@ internal object LDManager {
     internal fun initialize(app: Application, ldConfig: LDConfig) {
         val contextKind = ContextKind.of(LDContextKind.SERVICE)
         val context = LDContext.create(contextKind, LDContexts.ANDROID_CONTEXT)
-        client = LDClient.init(app, ldConfig, context, 0)
+        client = LDClient.init(app, ldConfig, context, 1)
```
djoksimo, line 52: is this the LD fix? Was this merged already?

dleis612: Yup this is the fix and it's going in!

PR #135, opened 2023-12-03
[FX-780] Fix empty state for PIN

dleis612:
## What
Fix the default PIN state to set `isValid` to `false`.

## Why
The PIN and PAN have different initial states

## Test Plan
Unit tests have been adjusted.


forage-android/src/test/java/com/joinforage/forage/android/core/element/state/ElementStateManagerTest.kt
```diff
@@ -7,7 +7,22 @@ import org.junit.Test
...
+    fun testGetStatePIN() {
+        val manager = PinElementStateManager.forEmptyInput()
+        val state = manager.getState()
+        assertThat(state.isFocused).isEqualTo(INITIAL_PIN_ELEMENT_STATE.isFocused)
+        assertThat(state.isBlurred).isEqualTo(INITIAL_PIN_ELEMENT_STATE.isBlurred)
+        assertThat(state.isEmpty).isEqualTo(INITIAL_PIN_ELEMENT_STATE.isEmpty)
+        assertThat(state.isValid).isEqualTo(INITIAL_PIN_ELEMENT_STATE.isValid)
+        assertThat(state.isComplete).isEqualTo(INITIAL_PIN_ELEMENT_STATE.isComplete)
+        assertThat(state.validationError).isEqualTo(INITIAL_PIN_ELEMENT_STATE.validationError)
+        // TODO: Compiler is very confused about accessing the state.details for the PIN
```
dleis612, line 19: @devinmorgan attempting to access the `details` field was a little surprising and the compiler is all confused.

devinmorgan: Hey Danny! Yeah agree that it throws off the compiler. I was thinking that this was actually OK if not even a good thing that trying to access `ElementState<PinDetails>` caused such headache because:
1. this would make it clear to merchant developers that details is currently meaningless for PINs and it is only meaningful for PANs (compiler has no issues with `ElementState<PanDetails>`)
2. The workaround for this is to do `state.details as Any?`, which didn't seem to be the too complicated

As you may have already gathered, the reason the compiler is upset about accessing `.details` on `ElementState<PinDetails>` is because `PinDetails` is currently a typealias for `Nothing?`. My thinking was that if we ever want to expose something meaningful on `PinDetails`, we can easily change the type to something substantive similar to `PanDetails`. Going from `PinDetails = Nothing?` --> `data class PinDetails(...)`  should not be a breaking change either because, presumably, nobody is going to be attempting to access `.details` for `ElementSTate<PinDetails` given that doing so would cause the compiler to yell at them + there is no meaningful values in there.

Let me know what you think about this and glad that you brought this up

PR #137, opened 2023-12-05
[FX-792] Add consumer-rules to Forage SDK

dleis612:
## What
Create a data class for the BT body parameters and add a directive to the `consumer-rules.pro` file so that people who use Forage don't obfuscate the object.

## Why
GoPuff (and probably others in the future) use ProGuard or a similar tool that obfuscates their code. Since they consume our SDK, this also obfuscates _OUR_ code. When this happens, the BT integration fails because the `card_number_token` and `pin` fields are changed to `a` and `b`. These names are important and must stay consistent.

## Test Plan
I have tested this change locally by running ProGuard on our sample app and replicating the issue. After adding the `consumer-rules.pro` directive, I re-ran the build process and can see that the code is not obfuscated. Then, I ran the release .apk file and was able to interact with BT.

### DIFF
--- a/forage-android/consumer-rules.pro
+++ b/forage-android/consumer-rules.pro
@@ -1,0 +1,3 @@
+-keepclassmembers class com.joinforage.forage.android.collect.ProxyRequestObject {
+    *;
+}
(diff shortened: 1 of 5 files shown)


devinmorgan (approved):
This PR is siiick!!! Nice work Danny!

PR #144, opened 2023-12-29
Create ForageVaultWrapper

dleis612:
## Test Plan

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/BTVaultWrapper.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/BTVaultWrapper.kt
@@ -11,6 +11,7 @@ import android.view.ViewGroup
 import android.widget.LinearLayout
 import com.basistheory.android.view.TextElement
 import com.basistheory.android.view.mask.ElementMask
+import com.google.android.material.textfield.TextInputEditText
 import com.joinforage.forage.android.core.element.state.PinElementStateManager
 import com.verygoodsecurity.vgscollect.widget.VGSEditText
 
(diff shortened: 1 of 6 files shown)


dleis612:
@devinmorgan it seems to me that the `onFocusChangeListener` seems to be working correctly, but I recall we had to add some special handling for this case in the `ForagePANEditText` class. Do you think [this](https://github.com/teamforage/forage-android-sdk/blob/27555338350553fc5dc65e325b9672f63509a3d3/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt#L192) is an issue here too?

devinmorgan (commented):
Nice work so far, Danny. My understanding is that you said there's still a bit of work you wanted to do in this PR so I'll wait to approve until you take it out of Draft. Cheers!

devinmorgan (approved):
Nice work on this PR Danny! Thanks for your patience with me in the review process. Exciting progress so far!

devinmorgan:
> @devinmorgan it seems to me that the `onFocusChangeListener` seems to be working correctly, but I recall we had to add some special handling for this case in the `ForagePANEditText` class. Do you think [this](https://github.com/teamforage/forage-android-sdk/blob/27555338350553fc5dc65e325b9672f63509a3d3/forage-android/src/main/java/com/joinforage/forage/android/ui/ForagePANEditText.kt#L192) is an issue here too?

@dleis612 - really thoughtful concern you bring up here!

You're right that we did some special handling in the past. I *suspect* that this will not be an issue in our case here because we're using a `android.widget.EditText` for as the PIN input. The problem in the past seems particular to `com.google.android.material.textfield.TextInputEditText`, which we were using in the `ForagePANEditText`.

👆 is just my best guess. In the worst case if we notice a similar problem, I think we can address it then. Lmk whether this is a productive response to your question though

PR #146, opened 2024-01-04
Add initial codepath for refund to pinCollector

dleis612:
## Test Plan

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
@@ -20,21 +20,13 @@ import com.verygoodsecurity.vgscollect.core.HTTPMethod
 import org.json.JSONException
 import java.util.*
 
+// IMPORTANT: Any changes to this object must be reflected in the consumer-rules.pro file
 internal data class ProxyRequestObject(val pin: TextElement, val card_number_token: String)
-
-internal data class POSProxyRequestObject(
-    val pin: TextElement,
-    val card_number_token: String,
-    val amount: String,
-    val reason: String,
-    val pos_terminal: Map<String, Any>,
-    val metadata: Map<String, Any>?
-)
 
 internal class BTPinCollector(
     private val pinForageEditText: ForagePINEditText,
     private val merchantAccount: String
-) : PinCollector() {
+) : PinCollector {
     private val logger = Log.getInstance()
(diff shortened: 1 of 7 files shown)


forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -22,10 +23,19 @@ import java.util.*
 
 internal data class ProxyRequestObject(val pin: TextElement, val card_number_token: String)
 
+internal data class POSProxyRequestObject(
+    val pin: TextElement,
+    val card_number_token: String,
+    val amount: String,
+    val reason: String,
+    val pos_terminal: Map<String, Any>,
```
djoksimo, line 31: can we make the type for `pos_terminal` stricter? This format is also used by the proxy `/balance` endpoint:

```
{
  pos_terminal: {
    provider_terminal_id: String
  }
}
```

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/collect/BTPinCollector.kt
```diff
@@ -22,10 +23,19 @@ import java.util.*
 
 internal data class ProxyRequestObject(val pin: TextElement, val card_number_token: String)
 
+internal data class POSProxyRequestObject(
```
djoksimo, line 26: Can we rename this to something "refund" specific. We'll also be sending POS-specific stuff for balance checks

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/network/ForageConstants.kt
```diff
@@ -18,6 +18,11 @@ internal object ForageConstants {
 
     object RequestBody {
         const val CARD_NUMBER_TOKEN = "card_number_token"
+        const val REASON = "reason"
+        const val METADATA = "metadata"
+        const val POS_TERMINAL = "pos_terminal"
+        const val POS_TERMINAL_ID = "pos_terminal_id"
```
djoksimo, line 24:
```suggestion
        const val POS_PROVIDER_TERMINAL_ID = "provider_terminal_id"
```

I think this is supposed to be `provider_terminal_id`

djoksimo (changes requested):

djoksimo:
Closing in favor of #159

PR #150, opened 2024-01-07
[FX-313] Make customerId optional

djoksimo:
## What

- Make `customerId` in `tokenizeEBTCard` method optional

## Test Plan

- ✅  Added unit tests for missing `customerId` case

## How

Can be released as-is

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ The ForageSDK exposes the following function to collect the EBT card number:
     data class TokenizeEBTCardParams(
         val foragePANEditText: ForagePANEditText,
         val customerId: String?,
-        val reusable: Boolean?
+        val reusable: Boolean = true
     )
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -263,8 +263,8 @@ The ForageSDK exposes the following function to collect the EBT card number:
 ```kotlin
     data class TokenizeEBTCardParams(
         val foragePANEditText: ForagePANEditText,
-        val customerId: String,
-        val reusable: Boolean
+        val customerId: String?,
+        val reusable: Boolean?
```
djoksimo, line 267: This was also made optional in #151

dleis612: Just a callout to be careful when merging this change in after #151 to make sure nothing is weird/overwritten

dleis612 (approved):

dleis612 (approved):

PR #152, opened 2024-01-09
Support generic vault request params

djoksimo:
## What

- Add support for generic `BaseVaultRequestParams`
- [x] Fix broken unit tests

## Why

- Easier to support pos-specific params while reusing existing facilities

## Test Plan

- ✅ Updated unit tests to use the new generic params

## How

Can be merged as-is and other branches for POS refund + balance will build on top of these changes

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
@@ -39,9 +39,9 @@ internal class VGSPinCollector(
         return vaultType
     }
 
-    override suspend fun <T : BaseVaultRequestParams>submitBalanceCheck(
+    override suspend fun submitBalanceCheck(
         paymentMethodRef: String,
-        vaultRequestParams: T
+        vaultRequestParams: BaseVaultRequestParams
     ): ForageApiResponse<String> = suspendCoroutine { continuation ->
         val logAttributes = mapOf(
             "merchant_ref" to merchantAccount,
@@ -128,9 +128,9 @@ internal class VGSPinCollector(
         vgsCollect.asyncSubmit(request)
     }
 
-    override suspend fun <T : BaseVaultRequestParams>submitPaymentCapture(
+    override suspend fun submitPaymentCapture(
         paymentRef: String,
-        vaultRequestParams: T
+        vaultRequestParams: BaseVaultRequestParams
     ): ForageApiResponse<String> = suspendCoroutine { continuation ->
         val logAttributes = mapOf(
             "merchant_ref" to merchantAccount,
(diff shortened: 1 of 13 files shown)


forage-android/src/main/java/com/joinforage/forage/android/network/data/BaseVaultRequestParams.kt
```diff
@@ -0,0 +1,16 @@
...
+) {
+    override fun equals(other: Any?): Boolean {
+        return other is BaseVaultRequestParams &&
+            other.cardNumberToken == cardNumberToken &&
+            other.encryptionKey == encryptionKey
+    }
+
+    override fun hashCode(): Int {
+        return cardNumberToken.hashCode() + encryptionKey.hashCode()
+    }
```
djoksimo, line 15: `data classes` do not support inheritance.

The overridden `equals` and `hashcode` are needed when `BaseVaultRequestParams` is used inside a `Map` (like in our unit tests, and possible other places in the future).

Open to learning about alternative workarounds.

forage-android/src/main/java/com/joinforage/forage/android/collect/VGSPinCollector.kt
```diff
@@ -349,12 +348,27 @@ internal class VGSPinCollector(
                 .create()
         }
 
-        private fun buildRequestBody(cardToken: String): HashMap<String, String> {
-            val body = HashMap<String, String>()
-            body[ForageConstants.RequestBody.CARD_NUMBER_TOKEN] = cardToken
+        internal fun buildRequestBody(requestBodyParams: BaseVaultRequestParams): HashMap<String, Any> {
```
dleis612, line 351: Nice :)

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -59,22 +59,24 @@ internal class CapturePaymentRepository(
...
+                        PaymentMethod.ModelMapper.from(response.data)
+                    ),
+                    encryptionKey = encryptionKey
+                )
             )
             else -> response
         }
     }
 
     private suspend fun submitPaymentCapture(
```
dleis612, line 73: I think this signature needs `<T : BaseVaultRequestParams>`

dleis612 (changes requested):

dleis612 (approved):

devinmorgan (approved):
PR looks good! One thing I wanted to clarify:

From what I can tell, this PR only updates `VGSPinCollector` shuttle the the pos related data into the proxy request and this PR does not update the `BTPinCollector` to do the same. Is that correct?

PR #154, opened 2024-01-10
[FX-769] Implement checkBalance for POS

djoksimo:
Introduced reusable factories, improved logging, got rid of T: BaseVaultRequestParams

## What

- [x] Implement `checkBalance` using `PosVaultRequestParams`
- [x] Add unit tests
- [x] Add reusable factories to reduce duplication inside `ForageSDK` and between `ForageSDK` and `ForageSDKInterface`
- [x] Make check balance server response + server mocking facilities reusable
- [x] Add log statements with unit tests for metrics logs + log statements

## Test Plan

- ✅ Added unit tests

## How

Can be released as-is with #152


forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -296,39 +204,84 @@ class ForageSDK : ForageSDKInterface {
...
+                measurement.setForageErrorCode(apiResponse.errors[0].code)
+            }
+            EventOutcome.FAILURE
+        } else {
+            EventOutcome.SUCCESS
+        }
+        measurement.setEventOutcome(outcome).logResult()
+    }
+
+    internal open class ServiceFactory(
```
dleis612, line 238: This is a great optimization!

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
```diff
@@ -127,7 +153,49 @@ class ForageTerminalSDK(
     override suspend fun checkBalance(
         params: CheckBalanceParams
     ): ForageApiResponse<String> {
-        return ForageApiResponse.Success("TODO")
+        val logger = createLogger()
+        val forageSdkInstance = forageSdk as ForageSDK
```
dleis612, line 157: Can we always safely cast this? I assume the type is `ForageSDKInterface` because we want to mock it in testing, but if we are going to straight cast it here should we just make the type `ForageSDK`?

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
```diff
@@ -127,7 +153,49 @@ class ForageTerminalSDK(
...
+        val balanceCheckService = serviceFactory.createCheckBalanceRepository(foragePinEditText)
+        val balanceResponse = balanceCheckService.posCheckBalance(
+            paymentMethodRef = paymentMethodRef,
+            posTerminalId = posTerminalId
+        )
+        forageSdkInstance.processApiResponseForMetrics(balanceResponse, measurement)
+
+        if (balanceResponse is ForageApiResponse.Failure) {
+            logger.e(
+                "[POS] checkBalance failed for PaymentMethod $paymentMethodRef on Terminal $posTerminalId ${balanceResponse.errors[0]}",
```
dleis612, line 190: Nit: Can you add a colon or something after the `$posTerminalId` to separate the error?

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
```diff
@@ -148,7 +216,28 @@ class ForageTerminalSDK(
...
+        val (_, paymentRef) = params
+
+        logger.i(
+            "[POS] Called capturePayment for Payment $paymentRef",
+            attributes = mapOf(
+                "payment_method_ref" to paymentRef
+            )
+        )
+
+        val captureResponse = forageSdk.capturePayment(params)
```
dleis612, line 229: Hmmmm we are going to get some extra log statements I think by calling the original function. It's not the biggest deal, but I would definitely consider it tech debt since we can expect a MASSIVE amount of volume some day with this.

forage-android/src/test/java/com/joinforage/forage/android/pos/ForageTerminalSDKTest.kt
```diff
@@ -78,14 +121,150 @@ class ForageTerminalSDKTest {
     }
 
     @Test
-    fun testCheckBalance() = runTest {
+    fun `POS EBT checkBalance should succeed`() = runTest {
```
dleis612, line 124: Potential nit?: These tests are massive. Is it super annoying to break them up somehow? If it is, then let's leave them as is.

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
djoksimo: ff835b192bb21e08bfed956cc3cf8b8801b70dba

I used `mockito` now instead of implementing the `ForageSDKInterface` to avoid the slightly more hazardous casting in `checkBalance`.

The original challenge was with mocking `ForageSDK` via inheritance as it is (and I think it should be) a final class

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
djoksimo: ff835b192bb21e08bfed956cc3cf8b8801b70dba

forage-android/src/test/java/com/joinforage/forage/android/pos/ForageTerminalSDKTest.kt
djoksimo: good callout, reduced duplication + bloating here: e8dc8c3f4b7844dbd76451d21250c0e03c5ba62d

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
```diff
@@ -127,7 +153,48 @@ class ForageTerminalSDK(
     override suspend fun checkBalance(
         params: CheckBalanceParams
     ): ForageApiResponse<String> {
-        return ForageApiResponse.Success("TODO")
+        val logger = createLogger()
+        val (foragePinEditText, paymentMethodRef) = params
+        val (merchantId, sessionToken) = forageSdk._getForageConfigOrThrow(foragePinEditText)
+
+        // TODO: replace Log.getInstance() with Log() in future PR
```
dleis612, line 160: Did you actually solve this TODO in this PR through the `createLogger()` function?

dleis612 (approved):

devinmorgan (approved):
Appreciate you cranking out the code so fast

PR #157, opened 2024-01-16
Introduce PollingService

djoksimo:
## What

de-dupe polling across payment capture and balance check

will be used for refund as well

## Why

- DRY up across balance check, capture and refund

## Test Plan

- ✅ unit tests cover polling path and pass

## How

Can be released as-is ; `PollingService` has further room for improvement, but this is a first step

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
@@ -11,6 +11,7 @@ import com.joinforage.forage.android.network.MessageStatusService
 import com.joinforage.forage.android.network.OkHttpClientBuilder
 import com.joinforage.forage.android.network.PaymentMethodService
 import com.joinforage.forage.android.network.PaymentService
+import com.joinforage.forage.android.network.PollingService
 import com.joinforage.forage.android.network.TokenizeCardService
 import com.joinforage.forage.android.network.data.CapturePaymentRepository
 import com.joinforage.forage.android.network.data.CheckBalanceRepository
@@ -83,13 +84,9 @@ class ForageSDK : ForageSDKInterface {
 
         // TODO: replace Log.getInstance() with Log() in future PR
         val logger = Log.getInstance()
-        logger.i(
-            "[HTTP] Tokenizing Payment Method",
-            attributes = mapOf(
-                "merchant_ref" to merchantId,
-                "customer_id" to customerId
-            )
-        )
+            .addAttribute("merchant_ref", merchantId)
+            .addAttribute("customer_id", customerId)
+        logger.i("[Android SDK] Tokenizing Payment Method")
 
         val tokenizeCardService = ServiceFactory(sessionToken, merchantId, logger)
             .createTokenizeCardService()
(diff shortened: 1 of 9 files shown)


forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -83,13 +84,9 @@ class ForageSDK : ForageSDKInterface {
...
-        logger.i(
-            "[HTTP] Tokenizing Payment Method",
-            attributes = mapOf(
-                "merchant_ref" to merchantId,
-                "customer_id" to customerId
-            )
-        )
+            .addAttribute("merchant_ref", merchantId)
+            .addAttribute("customer_id", customerId)
+        logger.i("[Android SDK] Tokenizing Payment Method")
```
dleis612, line 89: Why are we changing the prefix? We already know this log is from the Android SDK because it's logged under the `android-sdk` service.

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -195,13 +188,8 @@ class ForageSDK : ForageSDKInterface {
 
         // TODO: replace Log.getInstance() with Log() in future PR
         val logger = Log.getInstance()
-        logger.i(
-            "[HTTP] Submitting capture request for Payment $paymentRef",
```
dleis612, line 199: Log removed entirely.

forage-android/src/main/java/com/joinforage/forage/android/core/telemetry/Log.kt
```diff
@@ -90,27 +92,27 @@ internal interface Log {
                 }
                 return traceId as String
             }
+
+            override fun addAttribute(key: String, value: Any?): Log {
+                logger?.addAttribute(key, value)
```
dleis612, line 97: There is a chance that this function overwrites all attributes on the logger when it's called... I'm not certain but I feel like I vaguely remember that being a reason we only called it once per log. Can you confirm?

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
djoksimo: `HTTP` felt misleading given that the logs were coming from `ForageSDK`. I will update to `ForageSDK` to be inline with iOS log format to make it clearer that the logs are originating from the entrypoint of the SDK

forage-android/src/main/java/com/joinforage/forage/android/core/telemetry/Log.kt
djoksimo: Based on what I'm seeing here, it seems to cumulatively add context, like Datadog does in JS + iOS

[See logs](https://app.datadoghq.com/logs?query=service:android-sdk%20env:staging%20@payment_method_ref:45de62da41%20&cols=host,service&index=*&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1705421316134&to_ts=1705507716134&live=true)

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
djoksimo: Oops! Nice catch ; restored here: 0881b1a0a380b091cf28d9474ff4839f51cc4f2c

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
djoksimo: 0881b1a0a380b091cf28d9474ff4839f51cc4f2c

dleis612 (approved):

PR #129, opened 2023-11-14
WIP - need to implement loggers,  metrics, and tests

devinmorgan:
## Test Plan

- ✅ | ❌ I've added unit tests for this change. 
- ✅ | ❌ The reviewer should manually test the changes in this PR.

PR #158, opened 2024-01-17
Add Abstract PIN submitters

djoksimo:
## What

- Introduce
  - `AbstractVaultSubmitter`
  - `VgsPinSubmitter`
  - `BasisTheoryPinSubmitter`

Sought inspiration from @devinmorgan's design doc and his draft PR #129 🎉 

## Why

- Improve maintainability of PIN submission logic (across capturePayment, deferPaymentCapture, checkBalance, refundPayment, and so on.)
- Lay foundation for new refund PIN submission
- Lay foundation for implementing `ForageVaultSubmitter`

## Test Plan

- ✅  Plugged into `CheckBalanceRepository` locally and tests (that cover VGS responses) passed
- Will add unit tests for the new code in either this PR or a follow-up one!

## How

This PR can be released as-is

The PR will be followed by an "Add `RefundPaymentRepository`" PR that handles the polling using the new `PollingService` from #157 and the abstract submitter from this PR

- I will eventually refactor the existing facilities (balance, capture payment, defer payment capture); I have a WIP branch where I started to do this) danilo/wip-use-abstract-vault-submitters

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
@@ -142,11 +142,11 @@ internal abstract class AbstractVaultSubmitter<VaultResponse>(
             return UnknownErrorApiResponse
         }
 
-        val vaultApiError = toVaultErrorOrNull(vaultResponse)
-        if (vaultApiError != null) {
+        val vaultError = toVaultErrorOrNull(vaultResponse)
+        if (vaultError != null) {
             val forageErr = parseVaultError(vaultResponse)
             logger.e("[$vaultType] Received error from $vaultType: $forageErr")
-            return vaultApiError
+            return vaultError
         }
 
         val forageApiErrorResponse = toForageErrorOrNull(vaultResponse)
@@ -161,6 +161,7 @@ internal abstract class AbstractVaultSubmitter<VaultResponse>(
             logger.i("[$vaultType] Received successful response from $vaultType")
             return forageApiSuccess
         }
+        logger.e("[$vaultType] Received malformed response from $vaultType")
 
         return UnknownErrorApiResponse
     }
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
@@ -12,7 +12,6 @@ import com.joinforage.forage.android.network.model.ForageApiResponse
 import com.joinforage.forage.android.network.model.ForageError
 import com.joinforage.forage.android.network.model.UnknownErrorApiResponse
 import com.joinforage.forage.android.ui.ForagePINEditText
-import com.verygoodsecurity.vgscollect.core.HTTPMethod
 
 internal val IncompletePinError = ForageApiResponse.Failure.fromError(
     ForageError(400, "user_error", "Invalid EBT Card PIN entered. Please enter your 4-digit PIN.")
@@ -74,7 +73,7 @@ internal abstract class AbstractVaultSubmitter<VaultResponse>(
         )
         proxyResponseMonitor
             .setPath(path)
-            .setMethod(HTTPMethod.POST.toString())
+            .setMethod("POST")
             .start()
 
         val vaultProxyRequest = buildProxyRequest(
@@ -126,15 +125,14 @@ internal abstract class AbstractVaultSubmitter<VaultResponse>(
         .setHeader(ForageConstants.Headers.TRACE_ID, logger.getTraceIdValue())
         .setToken(vaultToken)
 
+    // PaymentMethod.card.token is in the comma-separated format <vgs-token>,<basis-theory-token>
     protected fun pickVaultTokenByIndex(paymentMethod: PaymentMethod, index: Int): String? {
         val tokensString = paymentMethod.card.token
         val tokensList = tokensString.split(TOKEN_DELIMITER)
 
-        // VGS is always the first token in the string
         val noTokenStoredInVault = tokensList.size <= index
         if (noTokenStoredInVault) return null
 
-        // grab BasisTheory token
         return tokensList[index]
     }
 
--- a/forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
@@ -1,0 +1,182 @@
+package com.joinforage.forage.android.collect
+
+import android.content.Context
+import com.joinforage.forage.android.VaultType
+import com.joinforage.forage.android.core.telemetry.Log
+import com.joinforage.forage.android.core.telemetry.UserAction
+import com.joinforage.forage.android.core.telemetry.VaultProxyResponseMonitor
+import com.joinforage.forage.android.model.EncryptionKeys
+import com.joinforage.forage.android.model.PaymentMethod
+import com.joinforage.forage.android.network.ForageConstants
+import com.joinforage.forage.android.network.model.ForageApiResponse
+import com.joinforage.forage.android.network.model.ForageError
+import com.joinforage.forage.android.network.model.UnknownErrorApiResponse
+import com.joinforage.forage.android.ui.ForagePINEditText
+import com.verygoodsecurity.vgscollect.core.HTTPMethod
+
+internal val IncompletePinError = ForageApiResponse.Failure.fromError(
+    ForageError(400, "user_error", "Invalid EBT Card PIN entered. Please enter your 4-digit PIN.")
+)
+
+internal interface VaultSubmitter {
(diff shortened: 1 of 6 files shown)


forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
```diff
@@ -0,0 +1,183 @@
...
+import com.joinforage.forage.android.network.model.ForageApiResponse
+import com.joinforage.forage.android.network.model.ForageError
+import com.joinforage.forage.android.network.model.UnknownErrorApiResponse
+import com.joinforage.forage.android.ui.ForagePINEditText
+import com.verygoodsecurity.vgscollect.core.HTTPMethod
+
+
+internal val IncompletePinError = ForageApiResponse.Failure.fromError(
+    ForageError(400, "user_error", "Invalid EBT Card PIN entered. Please enter your 4-digit PIN.")
+)
```
djoksimo, line 20: I know we have a dupe inside the `ForageConstants`, but I figured it made more sense for it to live here - where it is only used.

The dupe will be removed once the legacy PIN collectors become obsolete

forage-android/src/main/java/com/joinforage/forage/android/collect/VgsPinSubmitter.kt
```diff
@@ -0,0 +1,159 @@
...
+    ): ForageApiResponse<String> = suspendCoroutine { continuation ->
+        val vgsCollect = buildVGSCollect(context)
+        vgsCollect.bindView(foragePinEditText.getTextInputEditText())
+        val vgsInputField = foragePinEditText.getTextInputEditText()
+
+        vgsCollect.addOnResponseListeners(object : VgsCollectResponseListener {
+            override fun onResponse(response: VGSResponse?) {
+                // VGS wants us to call onDestroy, unlike Basis Theory
+                vgsCollect.onDestroy()
+                vgsInputField.setText("")
```
djoksimo, line 43: why do `ForagePINEditText.clearText()`([see code](https://github.com/teamforage/forage-android-sdk/blob/79dc8fd508d60fd5fbf5cc7b07a2a122dcc904d3/forage-android/src/main/java/com/joinforage/forage/android/ui/VGSVaultWrapper.kt#L120)) and the clearText inside here have different behaviors?

forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
```diff
@@ -0,0 +1,182 @@
...
+
+    protected fun pickVaultTokenByIndex(paymentMethod: PaymentMethod, index: Int): String? {
+        val tokensString = paymentMethod.card.token
+        val tokensList = tokensString.split(TOKEN_DELIMITER)
+
+        // VGS is always the first token in the string
+        val noTokenStoredInVault = tokensList.size <= index
+        if (noTokenStoredInVault) return null
+
+        // grab BasisTheory token
```
dleis612, line 137: Adjust comment this is a generic function.

forage-android/src/main/java/com/joinforage/forage/android/collect/BasisTheoryPinSubmitter.kt
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        return vaultToForageResponse(vaultResponse)
+    }
+
+    override fun parseVaultError(vaultResponse: BasisTheoryResponse): String {
+        return vaultResponse.exceptionOrNull().toString()
+    }
+
+    override fun toForageSuccessOrNull(vaultResponse: BasisTheoryResponse): ForageApiResponse.Success<String>? {
```
dleis612, line 72: I think this function should be identical to the `toForageErrorOrNull` function. We want to try to parse the response as a successful Forage response and return `null` if we don't get it.

forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
```diff
@@ -0,0 +1,182 @@
...
+import com.joinforage.forage.android.core.telemetry.UserAction
+import com.joinforage.forage.android.core.telemetry.VaultProxyResponseMonitor
+import com.joinforage.forage.android.model.EncryptionKeys
+import com.joinforage.forage.android.model.PaymentMethod
+import com.joinforage.forage.android.network.ForageConstants
+import com.joinforage.forage.android.network.model.ForageApiResponse
+import com.joinforage.forage.android.network.model.ForageError
+import com.joinforage.forage.android.network.model.UnknownErrorApiResponse
+import com.joinforage.forage.android.ui.ForagePINEditText
+import com.verygoodsecurity.vgscollect.core.HTTPMethod
```
dleis612, line 15: Depends on VGS?

forage-android/src/main/java/com/joinforage/forage/android/collect/VgsPinSubmitter.kt
dleis612: You can call `foragePinEditText.clearText()` or `vgsInputField.setText("")`. Both would work here.

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/collect/VgsPinSubmitter.kt
djoksimo: do we still have to call `vgsCollect.onDestroy()` @dleis612 ?

dleis612: Uhhhhh I'm not entirely sure I know what that call does actually.

djoksimo: According to the VGS method comment:


```kotlin
// Clear all information collected before by VGSCollect. Preferably call it inside onDestroy system's callback.
 fun onDestroy() {
        client.cancelAll()
        responseListeners.clear()
        storage.clear()
    }
```

dleis612: Seems like we should continue calling it.

forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
djoksimo: 28e4362a995f7e3c5f76225562da1dc2faead80c

forage-android/src/main/java/com/joinforage/forage/android/collect/AbstractVaultSubmitter.kt
djoksimo: 28e4362a995f7e3c5f76225562da1dc2faead80c

forage-android/src/main/java/com/joinforage/forage/android/collect/BasisTheoryPinSubmitter.kt
djoksimo: 6c72e226a80e5eeb580d56240707c80ec835c192

forage-android/src/main/java/com/joinforage/forage/android/collect/VgsPinSubmitter.kt
```diff
@@ -34,13 +34,11 @@ internal class VgsPinSubmitter(
...
         val vgsCollect = buildVGSCollect(context)
         vgsCollect.bindView(foragePinEditText.getTextInputEditText())
-        val vgsInputField = foragePinEditText.getTextInputEditText()
 
         vgsCollect.addOnResponseListeners(object : VgsCollectResponseListener {
             override fun onResponse(response: VGSResponse?) {
-                // VGS wants us to call onDestroy, unlike Basis Theory
+                // Clear all information collected before by VGSCollect
                 vgsCollect.onDestroy()
-                vgsInputField.setText("")
```
dleis612: Have you tried this out in practice? I assume the `setText()` function was there because the PIN was still visible after submission.

djoksimo: @dleis612 we call `forageEditText.clearText()` here https://github.com/teamforage/forage-android-sdk/pull/158/files#diff-66067141ba6138825bc49435c8ff29ee82f168779e92ce98ee46f10d7aafba8bR89

I kept the call to `onDestroy` inside that block -> where we have a reference to the `vgsCollect`

dleis612: Ok I see what you're saying. Just to confirm, we now call `clearText()` in a single place and the logic happens for both vaults when it does right?

dleis612 (approved):

forage-android/src/main/java/com/joinforage/forage/android/collect/VgsPinSubmitter.kt
djoksimo: yep!

PR #161, opened 2024-01-19
Fix POS refund params

djoksimo:
## What

- Use more logical param ordering
- Prefix POS-specific params with `Pos`

## How

- [ ] Need to update docs once the ordering is approved!

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/pos/PosMethodParams.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/pos/PosMethodParams.kt
@@ -24,24 +24,22 @@ data class PosTokenizeCardParams(
 /**
  * The [PosRefundPaymentParams] are only valid for in-store POS Terminal transactions.
  *
- * @property amount A positive decimal number that represents how much of the original
- * payment to refund in USD. Precision to the penny is supported.
- *
- * **The minimum amount that can be refunded is 0.01.**
- *
  * @property foragePinEditText The [ForagePINEditText] instance that collected the customer's PIN for the refund.
  * @property paymentRef A unique string identifier that refers to the Payment to be refunded
  * as it’s represented in Forage's database. Forage returns the paymentRef in response
  * to a successful capturePayment call or in response to a request from your server that creates a Payment.
- * @reason reason A string that describes why the Payment is to be refunded.
+ * @property amount A positive decimal number that represents how much of the original
+ * payment to refund in USD. Precision to the penny is supported.
+ * **The minimum amount that can be refunded is 0.01.**
+ * @property reason A string that describes why the Payment is to be refunded.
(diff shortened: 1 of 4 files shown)


forage-android/src/main/java/com/joinforage/forage/android/pos/PosMethodParams.kt
```diff
@@ -22,26 +22,26 @@ data class PosTokenizeCardParams(
...
-data class RefundPaymentParams(
+data class PosRefundPaymentParams(
     val amount: Float,
     val foragePinEditText: ForagePINEditText,
-    val metadata: Map<String, String>? = null,
     val paymentRef: String,
-    val reason: String
+    val reason: String,
+    val metadata: Map<String, String>? = null
 )
```
djoksimo, line 47: @dleis612 @devinmorgan lemme know if this ordering makes sense and is conventional!

dleis612: I assume alphabetical so looks good to me

dleis612 (approved):

forage-android/src/main/java/com/joinforage/forage/android/pos/PosMethodParams.kt
devinmorgan: As far as Kotlin / Android conventions, this ordering of params is perfectly. I didn't think I would have any thoughts on this but upon doing a second glance, how do you feel about the below order?
```kotlin
data class PosRefundPaymentParams(
    val foragePinEditText: ForagePINEditText,
    val paymentRef: String,
    val amount: Float,
    val reason: String,
    val metadata: Map<String, String>? = null
)
```

I'm looking at what we've been doing for [`ForageSDK`'s methods params](https://github.com/teamforage/forage-android-sdk/blob/a30ed74c6829c8aa935142825feeff1a87688c44/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKInterface.kt) and it seems like there's a soft convention that:
1. the `ForageElement` is the first param
2. any `ref`s are the second param

djoksimo: Will refactor to this format

devinmorgan: awesome!

djoksimo: 1b8a8cda2b22c231b0f7fffa063959e829526092

djoksimo: 1b8a8cda2b22c231b0f7fffa063959e829526092

devinmorgan (approved):
Looks great! Thanks for these changes!!

PR #164, opened 2024-01-22
[FX-816] Create a POS PIN capture screen

evanfreeze:
## What
Adds a screen to the balance inquiry flow to capture the PIN and perform the balance check. The results are dumped onto the screen for now, but there's a separate ticket to create a dedicated screen to show the balance. 

## Why
POS Cert App

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/ef399bf6-5943-45bd-9bed-6dea570aed4f

## How
Can be released as-is


sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
```diff
@@ -117,28 +124,50 @@ fun POSComposeApp(
...
                             })
                         }
                     },
                     onBackButtonClicked = { navController.popBackStack(POSScreen.BalanceInquiryScreen.name, inclusive = false) },
-                    tokenizedCardData = uiState.tokenizationResult?.toString(),
                     withPanElementReference = { panElement = it }
                 )
             }
+            composable(route = POSScreen.BIPINEntryScreen.name) {
+                PINEntryScreen(
```
devinmorgan, line 150: Woah, so you've basically made a reusable `PINEntryScreen` and `PANEntryScreen` components at this point? OMG we're going to FLY through these stories!!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/screens/ManualPANEntryScreen.kt
devinmorgan, line 1: my understanding is that we remove this because it navigates to the PIN screen after this so there was no reason to try and show this (we'll be on a different scene by then). Is that correct?

sample-app/src/main/java/com/joinforage/android/example/ui/pos/ui/ComposableForagePINEditText.kt
devinmorgan, line 1: this looks like a snippet that we may want to highlight in our README and/or docs to communicate with merchant developers how they can use the `Forage Android SDK` inside a Jetpack compose app. Nice work!!

devinmorgan (approved):
OMG take my money!!!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
evanfreeze: Yeah! So the `POSScreen.BIPINEntryScreen.name` part is just defining a PIN Entry route in the BI Flow, but the underlying Composable it's rendering (the `PINEntryScreen`) uses callbacks so it can be generic. We define what happens after tapping "Submit" or "Back" on either of those screens here in the NavHost, so in the Payment flow, for instance, we can re-use the same `PINEntryScreen` composable but give its buttons different behavior.

sample-app/src/main/java/com/joinforage/android/example/ui/pos/screens/ManualPANEntryScreen.kt
evanfreeze: Right, that was my thinking. I had it there to make sure the tokenization was working, but now we know it works because it moves to the PIN screen when it succeeds 👍🏻  Could probably have better error handling though, tbh.

sample-app/src/main/java/com/joinforage/android/example/ui/pos/ui/ComposableForagePINEditText.kt
evanfreeze: Yeah! Super cool stuff. I considered making a generic Composable like `ComposableForageElement` or something that could then be called with the element you want to use (PAN or PIN) since the underlying implementation of these two ComposableForageP(A|I)NEditText views are identical, but the types got kind of weird and it felt a big like an unnecessary optimization since there are only these two fields right now (PAN & PIN). Maybe down the road if we have more fields that's something we could play with.

PR #167, opened 2024-01-22
[FX-831] Create Screen to select the type of payment/purchase to perform

evanfreeze:
## What
Adds a screen where a user can select which type of payment/purchase to perform once they've chosen to create a payment/purchase from the main action screen

This also does some housekeeping on how screen files are organized since we're gonna have a lot of them

## Why
POS Cert app

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/00679f67-712f-4cb8-82e0-3d62632f7afd

## How
Can be merged as-is


devinmorgan (approved):
This man is FLYING!!! ✈️✈️✈️

PR #168, opened 2024-01-23
[FX-987] & [FX-988] Use ForageTerminalSDK for PAN tokenization & PIN submission instead of ForageSDK

evanfreeze:
## What
Switches to use `ForageTerminalSDK` for pan tokenization and pin submission instead of `ForageSDK`. This required hoisting the view model up to the Fragment (previously it was created by the POSComposeApp and thus not available in the Fragment) so we could store the terminal ID in the view model and UI state in the success/error callbacks when setting up the CPay SDK.

## Why
POS Cert App

## Demo
No visual changes other than you should see the terminal ID on the main POS screen if there is a terminal attached

## How
Needs verification on terminal, but assuming it works, can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
@@ -7,14 +7,17 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import androidx.fragment.app.Fragment
+import androidx.fragment.app.activityViewModels
 import com.joinforage.android.example.databinding.FragmentPosBinding
 import com.pos.sdk.DeviceManager
 import com.pos.sdk.DevicesFactory
 import com.pos.sdk.callback.ResultCallback
+import com.pos.sdk.sys.SystemDevice
 
 class POSFragment : Fragment() {
     private var _binding: FragmentPosBinding? = null
     private val binding get() = _binding!!
+    private val posViewModel by activityViewModels<POSViewModel>()
 
     override fun onCreateView(
         inflater: LayoutInflater,
@@ -27,7 +30,7 @@ class POSFragment : Fragment() {
         binding.composeView.apply {
             setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
             setContent {
-                POSComposeApp()
+                POSComposeApp(viewModel = posViewModel)
             }
         }
 
@@ -43,10 +46,13 @@ class POSFragment : Fragment() {
             object : ResultCallback<DeviceManager> {
                 override fun onFinish(deviceManager: DeviceManager) {
                     Log.i("CPay SDK", "DeviceManager created successfully")
+                    val terminalId = deviceManager.systemDevice.getSystemInfo(SystemDevice.SystemInfoType.IMEI)
+                    posViewModel.setTerminalId(terminalId)
                 }
 
                 override fun onError(i: Int, s: String) {
                     Log.i("CPay SDK", "Failed to create DeviceManager: $i,$s")
+                    posViewModel.setTerminalId("fakeDevTerminalId")
                 }
             }
         )
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
@@ -1,23 +1,16 @@
 package com.joinforage.android.example.ui.pos
 
 import android.os.Bundle
-import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.compose.ui.platform.ViewCompositionStrategy
 import androidx.fragment.app.Fragment
-import androidx.fragment.app.activityViewModels
 import com.joinforage.android.example.databinding.FragmentPosBinding
-import com.pos.sdk.DeviceManager
-import com.pos.sdk.DevicesFactory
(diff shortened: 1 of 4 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
```diff
@@ -43,10 +46,13 @@ class POSFragment : Fragment() {
             object : ResultCallback<DeviceManager> {
                 override fun onFinish(deviceManager: DeviceManager) {
                     Log.i("CPay SDK", "DeviceManager created successfully")
+                    val terminalId = deviceManager.systemDevice.getSystemInfo(SystemDevice.SystemInfoType.IMEI)
+                    posViewModel.setTerminalId(terminalId)
```
evanfreeze, line 50: I'm not super confident in this. I'm not sure what ID we want to use and whether this is the right one.  @devinmorgan would need to verify this on an actual terminal.

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
```diff
@@ -43,10 +46,13 @@ class POSFragment : Fragment() {
             object : ResultCallback<DeviceManager> {
                 override fun onFinish(deviceManager: DeviceManager) {
                     Log.i("CPay SDK", "DeviceManager created successfully")
+                    val terminalId = deviceManager.systemDevice.getSystemInfo(SystemDevice.SystemInfoType.SN)
+                    posViewModel.setTerminalId(terminalId)
```
devinmorgan: lmao - I feel like I just learned  shorthand way to do this from you. I was previously doing something longer

devinmorgan (commented):
Recapping our slack huddle. Changes to definitely keep:
1. ✅ using `ForageTerminalSDK` over `ForageSDK`
2. ✅ partitioning some state for the `terminalId` that can be accessed via the `ViewModel`. Up to your judgment as to whether that should live in `POSUIState` or on the `POSViewModel` directly
3. ✅ use a hard coded value for the `terminalId` for cases where the CPay SDK doesn't work (i.e. in Android Emulators)

Changes to walk back if that's OK with you:
1. hoisting of the `viewModel` into the `POSFragment` -- Devin misleadingly left some CPay code in the `POSFragment` and that should be moved anyway

sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/POSUIState.kt
```diff
@@ -4,6 +4,7 @@ import com.joinforage.android.example.network.model.tokenize.PaymentMethod
 import com.joinforage.android.example.ui.pos.MerchantDetailsState
 
 data class POSUIState(
+    val terminalId: String? = "tempDevTerminalId",
```
devinmorgan, line 7: this is great! love that the default value lives here too!

devinmorgan: do you care if it's `terminalId: String` (non-nullable) or is nullable preferred here?

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
```diff
@@ -81,11 +86,15 @@ class POSViewModel : ViewModel() {
...
 
-    fun checkEBTCardBalance(foragePINEditText: ForagePINEditText, paymentMethodRef: String, onSuccess: (response: BalanceCheck?) -> Unit) {
+    fun checkEBTCardBalance(foragePinEditText: ForagePINEditText, paymentMethodRef: String, onSuccess: (response: BalanceCheck?) -> Unit) {
+        if (terminalId == null) {
+            throw Error("Invalid POS Terminal ID")
+        }
+
         viewModelScope.launch {
-            val response = ForageSDK().checkBalance(
+            val response = ForageTerminalSDK(terminalId!!).checkBalance(
```
devinmorgan, line 95: siiiick

devinmorgan: if it ends up being OK that `terminalId` can be a `String` and not a `String?` then we could remove the `!!` but no pressure if there's an unfavorable tradeoff

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSFragment.kt
```diff
@@ -33,22 +29,4 @@ class POSFragment : Fragment() {
...
-                override fun onFinish(deviceManager: DeviceManager) {
-                    Log.i("CPay SDK", "DeviceManager created successfully")
-                }
-
-                override fun onError(i: Int, s: String) {
-                    Log.i("CPay SDK", "Failed to create DeviceManager: $i,$s")
-                }
-            }
-        )
-    }
```
devinmorgan, line 53: thanks for deleting this! This definitely should live somewhere else!

devinmorgan (approved):
Lmk whether you think `terminalId` can be `String` (non-null) instead of `String?` (nullable). If you think `String` (non-null) is OK, I can make that change in my PR or you can make that change here as well. I'll at least approve to give you flexibility to merge at will. Nice work!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/POSUIState.kt
evanfreeze: bringing this over from slack for posterity:
> Originally I wanted a way to differentiate between “we’re using the fallback value” and “something went wrong grabbing the value from CPay” because on an actual terminal if we didn’t grab the ID it might be a problem. But also maybe it doesn’t matter and we can just have it always be a string haha

PR #169, opened 2024-01-23
[FX-976] Add Vault pin submitter tests

djoksimo:
## What

- [x] Add unit tests for `VgsPinSubmitter.kt` and `BasisTheoryPinSubmitter.kt`
  - [x] This required including a test mockito dependency to use `anyOrNull` 
- [x] Rename `collect` package to `vault` package
- [x] Fix ForageError (from VGS proxy) unpacking ([more info](https://github.com/teamforage/forage-android-sdk/pull/169/files#r1464003479)); this issue exists in the `VGSPinCollector` as well
- [x] Fix vault + forage error handling in `BasisTheoryPinSubmitter.kt`

## How

Can serve as reference for writing `ForagePinSubmitter` tests @dleis612 

Will follow-up with PRs for `AbstractVaultSubmitter.kt` and `RefundPaymentRepository.kt`

### DIFF
--- a/forage-android/build.gradle
+++ b/forage-android/build.gradle
@@ -143,6 +143,7 @@ dependencies {
     // Use Mockito judiciously (mainly for mocking views)!
     // Opt for dependency injection and inheritance over Mockito
     testImplementation 'org.mockito:mockito-core:5.8.0'
+    testImplementation 'org.mockito.kotlin:mockito-kotlin:4.1.0'
 
     ksp 'com.squareup.moshi:moshi-kotlin-codegen:1.14.0'
 
--- a/forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
@@ -1,4 +1,4 @@
-package com.joinforage.forage.android.collect
+package com.joinforage.forage.android.vault
 
 import android.content.Context
 import com.basistheory.android.service.BasisTheoryElements
@@ -21,7 +21,8 @@ internal typealias BasisTheoryResponse = Result<Any?>
 internal class BasisTheoryPinSubmitter(
     context: Context,
     foragePinEditText: ForagePINEditText,
-    logger: Log
+    logger: Log,
+    private val buildVaultProvider: () -> BasisTheoryElements = { buildBt() }
 ) : AbstractVaultSubmitter<BasisTheoryResponse>(
     context = context,
     foragePinEditText = foragePinEditText,
@@ -48,7 +49,7 @@ internal class BasisTheoryPinSubmitter(
         .setHeader(ForageConstants.Headers.CONTENT_TYPE, "application/json")
 
     override suspend fun submitProxyRequest(vaultProxyRequest: VaultProxyRequest): ForageApiResponse<String> {
-        val bt = buildBt()
+        val bt = buildVaultProvider()
 
         val proxyRequest: ProxyRequest = ProxyRequest().apply {
             headers = vaultProxyRequest.headers
--- a/forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
@@ -3,6 +3,7 @@ package com.joinforage.forage.android.vault
 import android.content.Context
 import com.basistheory.android.service.BasisTheoryElements
 import com.basistheory.android.service.ProxyRequest
+import com.basistheory.android.service.getValue
 import com.joinforage.forage.android.VaultType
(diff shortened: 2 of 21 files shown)


forage-android/build.gradle
```diff
@@ -143,6 +143,7 @@ dependencies {
     // Use Mockito judiciously (mainly for mocking views)!
     // Opt for dependency injection and inheritance over Mockito
     testImplementation 'org.mockito:mockito-core:5.8.0'
+    testImplementation 'org.mockito.kotlin:mockito-kotlin:4.1.0'
```
djoksimo, line 146: Added to support `anyOrNull` mockito helper

forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
```diff
@@ -127,5 +134,30 @@ internal class BasisTheoryPinSubmitter(
...
+         * errors
+         *
+         * @return the BasisTheory JSON response body string if the response is a BasisTheory error,
+         * or null if it is not
+         */
+        private fun getBasisTheoryExceptionBody(vaultResponse: BasisTheoryResponse): String? {
+            val rawBasisTheoryError = vaultResponse.exceptionOrNull()?.message ?: return null
+            val bodyRegex = "HTTP response body: (.+?)\\n".toRegex()
+            return bodyRegex.find(rawBasisTheoryError)?.groupValues?.get(1)
+        }
```
djoksimo, line 151: I reached out to the Basis Theory team about this limitation with ApiException.

It is still safe to release these changes as-is

forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
```diff
@@ -66,51 +67,57 @@ internal class BasisTheoryPinSubmitter(
...
+
+            val forageApiError = ForageApiError.ForageApiErrorMapper.from(basisTheoryExceptionBody)
+            val firstError = forageApiError.errors[0]
+
+            return ForageApiResponse.Failure.fromError(
+                ForageError(matchedStatusCode, firstError.code, firstError.message)
             )
-        } catch (_: JSONException) {
-            // if we throw when trying to extract the ForageApiError,
+        } catch (_: Exception) {
```
dleis612, line 100: The reason that we have always try-catched these functions is because the pattern you see in a lot of places where we try to parse to JSON (ex. `ForageApiError.ForageApiErrorMapper.from(basisTheoryExceptionBody)`) throws if the string doesn't conform.

Should we have two catch statements here? The first for `JSONException` and the second for all other exceptions so that we can handle them more carefully?

forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
```diff
@@ -66,51 +67,57 @@ internal class BasisTheoryPinSubmitter(
...
         if (vaultResponse.isSuccess) return null
 
-        // if the result is a Failure, we need to fail gracefully and
-        // return a ForageApiError
-        return UnknownErrorApiResponse
+        try {
+            // try to parse as an error from Basis Theory, indicated by the presence
+            // of a "proxy_error" in the raw response body.
+            val basisTheoryExceptionBody = getBasisTheoryExceptionBody(vaultResponse)
+            if (basisTheoryExceptionBody!!.contains("proxy_error")) {
```
dleis612, line 114: Maybe we should add some logging around this? I'm not sure I'm 100% comfortable with all this string parsing it seems like it could be a recipe for uncaught exceptions.

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
djoksimo: The reason I use a regular `Exception` here is because the regex parsing may throw a `NullPointerException` (or maybe even an out-of-bounds exception) and the ForageApiErrorMapper may throw a `JSONException`

In any case, an exception in this block means that the vault response doesn't conform to `ForageApiError`

So instead of:

```kotlin
catch (_: JSONException) {
  null
} catch (_: NullPointerException) {
  null
} catch (_: IndexOutOfBoundsException) {
 null
}
```

I went with a more shorthand approach (`catch (_: Exception) {}`)

dleis612: You're right that the outcome would be your proposal above. Let's just go for it.

forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
dleis612: Well put, I'm on board.

dleis612 (approved):

PR #171, opened 2024-01-24
House ForageConfig info in POSUIState

devinmorgan:
[House ForageConfig info](https://github.com/teamforage/forage-android-sdk/pull/171/commits/ff3579d5ee96cddcb1024686707d9a826f59474d)

I think the `POSUIState` could be a good home for the `merchantId`
and the `sessionToken`, which together constitute the
`ForageConfig` info. The wins that I see are:
1. easier to work with the `forageConfig` because a single method
    / getting on `POSUIState` will house the `forageConfig`
    Previously, to pass `forageConfig` to the `ForageTerminalSDK`,
    we needed to assemble the `ForageConfig` object ourselves
2. now there is a single file that developers can copy/paste the
    the `sessionToken` and `merchantId` into instead of needing
    to change values across two files

PR #172, opened 2024-01-24
Move terminalId to POSComposeApp

devinmorgan:
[Move terminalId to POSComposeApp](https://github.com/teamforage/forage-android-sdk/pull/172/commits/0ea4a6af8ef9cad2db8b8267db2992becb803a40) 

`terminalId` is similar to `sessionToken` and `merchantId` in that
it likely only gets set once with the initial fetch of merchant
details and then, afterwards, will not need to be set again. It
could follow from this that `terminalId` should live alongside
`merchantId` and `sessionToken` inside `POSUIState`.

I am not opposed to `terminalId` ultimately living in
`POSUIState`. However, reading the `terminalId` from the K9SDK
is an asynchronous process because we need to wait for the
K9SDK to finish initializing before we can read the `terminalId`
value.

It was not readily apparent to me how we would request the K9SDK
to initialize once and only once from within `POSComposeApp` and
then after the async initialization has completed, read the value
of `terminalId`. I do believe it's possible to do this and once we
figure this out, this would probably be the better way to go.

In the interest of speed, I opted for using `remember` and then
reading the value of `k9SDK.terminalId` as needed throughout
`POSComposeApp`

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
@@ -16,9 +16,11 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
@@ -29,6 +31,7 @@ import androidx.navigation.compose.composable
 import androidx.navigation.compose.currentBackStackEntryAsState
 import androidx.navigation.compose.rememberNavController
 import com.joinforage.android.example.R
+import com.joinforage.android.example.pos.k9sdk.K9SDK
 import com.joinforage.android.example.ui.pos.screens.ActionSelectionScreen
 import com.joinforage.android.example.ui.pos.screens.MerchantSetupScreen
 import com.joinforage.android.example.ui.pos.screens.balance.BalanceInquiryScreen
@@ -68,6 +71,11 @@ fun POSComposeApp(
         mutableStateOf(null)
     }
 
+    val context = LocalContext.current
+    val k9SDK by remember {
+        mutableStateOf(K9SDK().init(context))
+    }
+
     Scaffold(
         topBar = {
             TopAppBar(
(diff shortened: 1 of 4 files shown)


devinmorgan:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/172?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `main`
  * **PR #171** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #172** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/172?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
      * **PR #173** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/173?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/172?utm_source=stack-comment).

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
```diff
@@ -68,6 +71,11 @@ fun POSComposeApp(
         mutableStateOf(null)
     }
 
+    val context = LocalContext.current
+    val k9SDK by remember {
```
evanfreeze, line 75: `remember` will persist across "recompositions" (re-renders in React-speak) but `rememberSaveable` will persist across recompositions _and_ configuration changes. I'm don't think it matters either way for this, just calling it out in case it does.

evanfreeze (approved):
I like this approach of encapsulating all of the device-specific stuff into its own class. I agree about `terminalId` not really needing to be in the state or view model. I think I initially put it in the view model just so we didn't have to pass it into the VM's methods, but that's not a valid reason 😅 We have function parameters, might as well use 'em!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
devinmorgan: I tried to use `rememberSaveable` and the app crashed. I thiiiiiink it had to do with the fact that `K9SDK` is not serializable. I didn't bother to investigate a workaround at the time but I completely agree that `rememberSaveable` is better. I'll make a linear ticket to track this tech debt because I bet this will cause a bug at some point in the future. Great callout!

evanfreeze: Ahh I got ya, makes sense. Tbh I'm not sure this app will undergo many configuration changes anyway 😅

PR #180, opened 2024-01-25
[FX-748] Story: EBT cash withdrawal flow

evanfreeze:
## What
Adds support for initiating an EBT Cash Withdrawal transaction to the POS Cert App. _Note:_ This currently doesn't succeed due to backend changes needed, but the call appears to be correct based on the Linear ticket.

## Why
https://linear.app/joinforage/issue/FX-748/story-cash-withdrawal-flow

## Test Plan
- ❌ Just getting basic functionality in place
- ❌ We'll need to verify this succeeds once the backend changes are in place

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/d8c0895c-742f-4648-8912-2cc716f45955

## How
Can be merged as-is since the call failing doesn't affect the other flows and we catch the error


devinmorgan (approved):
Looks good!

PR #181, opened 2024-01-25
[FX-749] Story: Cash purchase with cash back flow

evanfreeze:
## What
Adds the ability to initiate a cash purchase with cash back transaction in the POS Cert App. _Note:_ This currently doesn't succeed due to backend changes needed, but the call appears to be correct based on the Linear ticket.

## Why
https://linear.app/joinforage/issue/FX-749/story-cash-purchase-with-cash-back-flow

## Test Plan
- ❌ Just getting basic functionality in place
- ❌ We'll need to verify this succeeds once the backend changes are in place

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/d5f7c966-33bb-4206-9901-658c36c91b45

## How
Can be merged as-is since the call failing doesn't affect the other flows and we catch the error

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/PosPaymentRequest.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/PosPaymentRequest.kt
@@ -10,7 +10,9 @@ data class PosPaymentRequest(
     @Json(name = "payment_method") val paymentMethodRef: String,
     val description: String,
     @Json(name = "pos_terminal") val posTerminal: PosTerminal,
-    val metadata: Map<String, String>
+    val metadata: Map<String, String>,
+    @Json(name = "transaction_type") val transactionType: String? = null,
+    @Json(name = "cash_back_amount") val cashBackAmount: Double? = null
 ) {
     companion object {
         fun forSnapPayment(snapAmount: Double, terminalId: String) = PosPaymentRequest(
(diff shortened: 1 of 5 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/PosPaymentRequest.kt
```diff
@@ -10,7 +10,9 @@ data class PosPaymentRequest(
     @Json(name = "payment_method") val paymentMethodRef: String,
     val description: String,
     @Json(name = "pos_terminal") val posTerminal: PosTerminal,
-    val metadata: Map<String, String>
+    val metadata: Map<String, String>,
+    @Json(name = "transaction_type") val transactionType: String? = null,
+    @Json(name = "cash_back_amount") val cashBackAmount: Double? = null
```
devinmorgan, line 15: damn these `@Json` tags with the `@JsonClass` seem really convenient!

devinmorgan (commented):
amazing!!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/PosPaymentRequest.kt
evanfreeze: right? so nice to send a snake case key name in json but just get to use it as camel case in kotlin

devinmorgan: I feel like we were chatting about wanting to do this in our repos at some point to keep the responses strongly typed but I think at the time it felt like it could be a "big deal".....but this....this feels not so bad...almost easy??

PR #182, opened 2024-01-25
[TW-397] Update comments for ForageElement

kimberleehowley:
## What

This PR adds comments to the ForageElement methods. 

## Why

The goal of these comments are to better document the SDK so that it is easier to use. 

## Test Plan

- ❌ |  No unit tests are needed for strictly comment changes. 
- ✅ | The reviewer should manually test the changes in this PR: 

Checkout my branch, and then run: 

```
./gradlew dokkaHtml 
npx http-serve reference-docs
```

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
@@ -126,7 +126,7 @@ interface ForageElement<T : ElementState> {
     fun setOnBlurEventListener(l: SimpleElementListener)
 
     /**
-     * Sets an event listener to be fired on ForageElement change events.
+     * Sets an event listener to be fired when the text inside the ForageElement input field changes.
      *
      * @param l The [StatefulElementListener] to be fired on change events.
      */
--- a/forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
@@ -30,15 +30,17 @@ data class ForageConfig(
 )
 
 /**
- * The interface that all Forage UI Components adhere to. For example,
- * both the ForagePANEditText and the ForagePINEditText satisfy
- * ForageElement
+ * The interface that defines methods for configuring and interacting with a [ForageElement].
+ * Both [ForagePANEditText] and [ForagePINEditText] adhere to this interface.
+ *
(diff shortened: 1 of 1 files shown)


dleis612 (approved):

forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
```diff
@@ -30,34 +30,105 @@ data class ForageConfig(
...
+     * @param boxStrokeWidth The scaled pixel size.
+     */
     fun setBoxStrokeWidth(boxStrokeWidth: Int)
+
+    /**
+     * Sets the border thickness of the input field when the field is in focus state.
+     *
+     * @param boxStrokeWidth The scaled pixel size.
+     */
     fun setBoxStrokeWidthFocused(boxStrokeWidth: Int)
```
djoksimo, line 105: @devinmorgan to verify these descriptions

forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
```diff
@@ -30,34 +30,105 @@ data class ForageConfig(
...
+
+    /**
+     * Sets an event listener to be fired when the ForageElement is blurred.
+     *
+     * @param l The [SimpleElementListener] to be fired on blur events.
+     */
     fun setOnBlurEventListener(l: SimpleElementListener)
+
+    /**
+     * Sets an event listener to be fired on ForageElement change events.
```
djoksimo, line 129: Can we specify when change events are fired; 

i.e. when the text inside the input field changes.

djoksimo (changes requested):
Nice stuff! Left a comment about the change listener

forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
kimberleehowley: Sure thing - I just pushed up this change!

djoksimo (approved):

forage-android/src/main/java/com/joinforage/forage/android/ui/ForageElement.kt
kimberleehowley: @devinmorgan Hello! I'll keep this PR open until Monday. If I don't read any feedback from you before then, I'll assume with Danilo and Danny's ✅  I'm good-to-go. Thank you!

PR #183, opened 2024-01-25
Add error messages in Cert app, use HOC for shared layout in screens

evanfreeze:
## What
Updates screens to use the `ScreenWithBottomRow` HOC to share common layout. Shows error messages in the UI when network requests fail or return error responses.

## Why
POS Cert app

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/3aa735a8-4f20-49ca-9e3c-03ffaccd1c08

## How
Can be merged as-is


sample-app/src/main/java/com/joinforage/android/example/ui/pos/data/POSUIState.kt
devinmorgan, line 1: These changes are awesome! Love that you grouped the related states together too. Makes it very easy to read!

devinmorgan (approved):
Great work Evan! Thanks for your patience!!

PR #184, opened 2024-01-25
[FX-996] Migrate PinCollector to AbstractVaultSubmitter

djoksimo:
## What

- refactor usage of `PinCollector` to `AbstractVaultSubmitter`
- use more linear flow of method calls for `checkBalance`, `capturePayment`, and `deferPaymentCapture`
- update tests to be compatible with vault submitter
- simplify `MockVaultSubmitter`
- remove unused `PinCollector` code + tests

## Test Plan

- ✅  Updated tests

## How

Can be released as-is. Should be merged to unblock ForageVaultSubmitter


forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
```diff
@@ -18,6 +18,8 @@ import com.joinforage.forage.android.ui.ForagePINEditText
 
 internal typealias BasisTheoryResponse = Result<Any?>
 
+internal data class ProxyRequestObject(val pin: TextElement, val card_number_token: String)
```
dleis612, line 21: We need to update the `consumer-rules.pro` file to reflect that new location of this object. Can you also add a comment about how important it is that we remember to do that in the future?

This is a very unsustainable pattern we have going here haha maybe we move this declaration to its own file with comments so we don't have to worry about it again?

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitter.kt
djoksimo: Good call!

Handled here: b8a15f15c6332da2479bc740f48ddf46dfbcc4f0

dleis612 (approved):

PR #187, opened 2024-01-29
[FX-746] Story: Return or refund a transaction

evanfreeze:
## What
Adds a flow to support refunding payments. Here's [an example of a refund created through the android emulator](https://api.dev.joinforage.app/admin/api/refund/2407/change/) in dev.

## Why
https://linear.app/joinforage/issue/FX-746/story-snap-cash-return-refund-flow

## Test Plan
- ❌ Early scaffolding of basic functionality
- ❌ Demo video should be sufficient

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/85ce4ac7-429c-4d88-bda9-f2ba2fce7d14

## How
Can be merged as-is


devinmorgan (approved):
Excellent! Excellent! Excellent!

PR #194, opened 2024-02-01
Introduce setPosForageConfig

djoksimo:
## What

- Introduce `ForageElement#setPosForageConfig` and `PosForageConfig` ; helps us use custom logic for loading vault provider-related elements.
- Added comments and validation in case the merchant uses the wrong method.
- gitignore `.idea` directory
- fix comments (for merchant ID, reference `ForageElement` instead of `AbstractForageElement`)

## Why

- [See design proposal doc](https://www.notion.so/joinforage/Handling-SDK-vault-routing-for-Toast-integration-f4b8aa60776640e9a7de013f48c13ea3#ca06e1abcf9b47f0a1c1ec5d7f3d8b9e)

## Test plan

- flipped flags in launchdarkly and tested against sample app
- wrote unit tests
- checked logs

## How

- PATCH-level change
- Will eventually be followed by PR for ensuring POS transactions go through rosetta
- ⚠️  @devinmorgan the certification app should be updated to use `setPosForageConfig` instead of `setForageConfig`

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
@@ -7,11 +7,14 @@ import com.joinforage.forage.android.ForageConfigNotSetException
 import com.joinforage.forage.android.ForageSDK
 import com.joinforage.forage.android.ForageSDKInterface
 import com.joinforage.forage.android.TokenizeEBTCardParams
+import com.joinforage.forage.android.VaultType
 import com.joinforage.forage.android.core.telemetry.CustomerPerceivedResponseMonitor
 import com.joinforage.forage.android.core.telemetry.Log
 import com.joinforage.forage.android.core.telemetry.UserAction
 import com.joinforage.forage.android.network.model.ForageApiResponse
+import com.joinforage.forage.android.network.model.UnknownErrorApiResponse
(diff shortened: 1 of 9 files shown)


forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
```diff
@@ -300,6 +252,79 @@ class ForageTerminalSDK(
...
+        return refund
+    }
+
+    private fun isIllegalVaultExceptionOrNull(
+        foragePinEditText: ForagePINEditText,
+        logger: Log
+    ): ForageApiResponse<String>? {
+        if (foragePinEditText.getVaultType() != VaultType.VGS_VAULT_TYPE) {
+            logger.e("[POS] checkBalance failed on Terminal $posTerminalId because the vault type is not VGS")
+            return UnknownErrorApiResponse
```
dleis612, line 323: Thoughts on returning back a new developer friendly error to push them to use the POS params?

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/pos/ForageTerminalSDK.kt
djoksimo: Great suggestion!

See diff here: 7de1243da26a20e9371654523f21bad3f9a70031

dleis612 (approved):

PR #198, opened 2024-02-02
Add session token constructor arg for key management and update README

ForageRob:
## What

Title

## Why

This param is needed for key management which is still under development. I don't want the extra constructor arg to be a surprise when we turn that work over, in case it is a significant flow change.

## Test Plan

- ❌ I've not added unit tests for this change. We will do that with the key management work itself

## How

Should be included in the next release

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,15 +300,16 @@ You can then use the instance to call methods that perform payment operations. R
 
 `ForageTerminalSDK()` is the entry point to the Forage Terminal SDK. A `ForageTerminalSDK` instance interacts with the Forage API.
 
-To create a `ForageTerminalSDK` instance, call the constructor, passing the `posTerminalId` as the only parameter, as in the following snippet:
+To create a `ForageTerminalSDK` instance, call the constructor, passing the `posTerminalId` and a `sessionToken` as the only parameters, as in the following snippet:
 
 ```kotlin
-val forage = ForageTerminalSDK(posTerminalId)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -300,15 +300,16 @@ You can then use the instance to call methods that perform payment operations. R
...
 
 ```kotlin
-val forage = ForageTerminalSDK(posTerminalId)
+val forage = ForageTerminalSDK(posTerminalId, sessionToken)
 ```
 
 #### `ForageTerminalSDK()` parameters
 
 - `posTerminalId` (_required_): A string that uniquely identifies the POS Terminal used for a transaction.
+- `sessionToken` (_required_): A session token which will authenticate the POS Terminal against Forage's key management service.
```
kimberleehowley, line 312: Thank you so much for doing this! You didn't miss anything. I'm just suggesting super nits that to this line 312 that: 

- Mirror the description Danilo and I wrote in Kdoc comments this week 
- Note that it's short-lived, link to the docs 

A short-lived token that authenticates the POS Terminal against Forage's key management service.  To create one, send a server-side `POST` from your backend to the [`/session_token/`](https://docs.joinforage.app/reference/create-session-token) endpoint.

kimberleehowley (approved):
I left the nittiest nits, thank you so much!

devinmorgan:
We added this PR inside of the `forage-pos-android-sdk` instead so I'm closing this

PR #200, opened 2024-02-05
[FX-1033] Add deferred refund endpoint to terminal SDK

dleis612:
## What
Add the deferred refund endpoint to the Android SDK for Toast to use.

## Test Plan

## How
Can be released as is.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKInterface.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKInterface.kt
@@ -167,3 +167,22 @@ data class DeferPaymentCaptureParams(
     val foragePinEditText: ForagePINEditText,
     val paymentRef: String
 )
+
+/**
+ * A model that represents the parameters that Forage requires to collect a card PIN and defer
+ * the refund of the payment to the server.
+ * [PosDeferPaymentRefundParams] are passed to the
+ * [deferPaymentRefund][com.joinforage.forage.android.pos.ForageTerminalSDK.deferPaymentRefund] function.
+ *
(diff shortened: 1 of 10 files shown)


devinmorgan (commented):
Had a look through it and it seems fine to me! Let me know when you need a +1 and I'll mark it as approved!

evanfreeze:
Verified this on staging by checking out the branch and roughing out some UI to trigger it in the cert app — it works 👍🏻

forage-android/src/main/java/com/joinforage/forage/android/ForageSDKInterface.kt
```diff
@@ -167,3 +167,22 @@ data class DeferPaymentCaptureParams(
...
+ *
+ * @property foragePinEditText A reference to a [ForagePINEditText] instance.
+ * [setForageConfig][com.joinforage.forage.android.ui.ForageElement.setForageConfig] must
+ * be called on the instance before it can be passed.
+ * @property paymentRef A unique string identifier for a previously created
+ * [`Payment`](https://docs.joinforage.app/reference/payments) in Forage's
+ * database, returned by the
+ * [Create a `Payment`](https://docs.joinforage.app/reference/create-a-payment) endpoint.
+ */
+data class DeferPosPaymentRefundParams(
```
djoksimo, line 185: the other public POS-specific `data class`-es are prefixed with `Pos`. 
```suggestion
data class PosDeferPaymentRefundParams(
```

djoksimo (changes requested):
Looks great! Left one comment about using the `Pos` prefix, but everything else LGTM.

PR #201, opened 2024-02-05
[FX-1028] Pull latest forage pos config inside data-fetching methods

evanfreeze:
## What
Pulls the latest POS config for retrofit api methods.

## Why
Since were initializing `api` using `by lazy`, the first time one of these methods gets called (which would almost certainly `getMerchantInfo` based on the app flow), we would initialize the `PosApiService` with whatever the value of `uiState.value.posForageConfig` was at that moment. Since `by lazy` caches its value once it runs, we'd never update the retrofit client creation. That wouldn't theoretically be a problem, but since [we inject the merchant ID in the headers](https://github.com/teamforage/forage-android-sdk/blob/main/sample-app/src/main/java/com/joinforage/android/example/ui/pos/network/PosApiService.kt#L53) when initializing the retrofit client, it would always keep the same merchant ID from first call.

## Test Plan
- ✅  I manually tested this based on the description in the [linear ticket](https://linear.app/joinforage/issue/FX-1028/bug-app-uses-old-fns-number-merchantid-for-requests-despite-user) and verified the behavior seems to be working correctly. 

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/593659c1-b1e7-4b10-9ccc-4d225e911b3b

## How
Can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
@@ -44,12 +44,9 @@ sealed interface MerchantDetailsState {
 class POSViewModel : ViewModel() {
     private val _uiState = MutableStateFlow(POSUIState())
     val uiState: StateFlow<POSUIState> = _uiState.asStateFlow()
-    private val api: PosApiService by lazy {
-        // lazy because we may not have the merchantId
-        // of the forageConfig when a ViewModel instance
-        // is created but we certainly will by the time
-        // we ever use any methods of the api
-        PosApiService.from(uiState.value.posForageConfig)
+
+    private fun getForageRetrofitApi(): PosApiService {
+        return PosApiService.from(uiState.value.posForageConfig)
     }
 
     fun setMerchantId(merchantId: String, onSuccess: () -> Unit) {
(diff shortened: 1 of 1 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
```diff
@@ -44,12 +44,9 @@ sealed interface MerchantDetailsState {
...
-    private val api: PosApiService by lazy {
-        // lazy because we may not have the merchantId
-        // of the forageConfig when a ViewModel instance
-        // is created but we certainly will by the time
-        // we ever use any methods of the api
-        PosApiService.from(uiState.value.posForageConfig)
+
+    private fun getForageRetrofitApi(): PosApiService {
+        return PosApiService.from(uiState.value.posForageConfig)
     }
```
devinmorgan, line 50:
```suggestion
    private val api
        get() = PosApiService.from(uiState.value.posForageConfig)
```
Nice fix! I'm embarrassed that I introduced this bug without evening realizing it. Thanks for identifying the problem and fixing 💛 .

Throwing out this suggestion if you want because it maaaaaay be a more Kotlin-y way to do this. I've been getting the impression that kotlin seems to opt for .accessors over `getters` / `setters`. But I haven't taken the time to read through conventions so I could be full of sh*t. I'm going to approve the PR since this fix works!

devinmorgan (approved):
Thanks for identifying and fixing! Sorry I introduced this bug in the first place!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
evanfreeze: Nice! Updated to this approach in [d643a4f](https://github.com/teamforage/forage-android-sdk/pull/201/commits/d643a4f23b0a59a82c375196a4b5452f412f81f9)

PR #204, opened 2024-02-06
[FX-1030] Display the name of the current flow throughout

evanfreeze:
## What
Adds more descriptive names for each screen, also enables overwriting the page title in code as necessary so we can specify which payment flow we're in despite them all sharing the same PAN & PIN screen (which would thus require them to have the same title, normally)

Also adds the Payment Method Ref to the PIN entry screen

## Why
https://linear.app/joinforage/issue/FX-1030/feature-display-the-name-of-the-current-flow-throughout-all-subscenes

## Demo
CleanShot 2024-02-06 at 15 42 01@2x

https://github.com/teamforage/forage-android-sdk/assets/11556475/8bf280de-59a1-4d32-9505-f55d4f89de72

## How
Can be merged as-is


sample-app/src/main/java/com/joinforage/android/example/ui/pos/screens/shared/PinEntryScreen.kt
```diff
@@ -22,6 +30,13 @@ fun PINEntryScreen(
     ScreenWithBottomRow(
         mainContent = {
             if (paymentMethodRef != null) {
+                Card {
+                    Column(modifier = Modifier.padding(16.dp)) {
+                        Text("Payment Method", fontWeight = FontWeight.SemiBold)
+                        Text("Ref: " + paymentMethodRef)
```
devinmorgan, line 36: These details are very helpful! Can we also display the `Refund.ref` if its a `Refund` flow? That should not block this PR but I wonder if that could be helpful to show `Payment.ref` / `Refund.ref` defpending on the flow

devinmorgan (approved):
Nice work Evan!!

PR #206, opened 2024-02-08
[FX-1052] Fix purchase with cashback throwing a 400 error

evanfreeze:
## What
Fixes an issue with the purchase + cashback flow throwing a 400 error. It looks like the `ebtCashAmount` and `cashBackAmounts` are coming into the helper method that generates the payment as strings, then we're trying to add them with `+`, which produces a result that the backend can't coerce to a number. This changes the logic to cast each item to a double, do the math to get the total, then cast that total _back_ to a string to send to the backend.

## Why
https://linear.app/joinforage/issue/FX-1052/bug-cash-purchase-with-cashback-gives-400-error

## Test Plan
Manually tested the flow in the android emulator

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/834ab196-5609-45ab-8190-81369592fdbd

## How
Can be merged as-is


devinmorgan (approved):
I feel like simply fixes like this always hide the hours of tears behind them. Either way, nice work!

PR #207, opened 2024-02-08
[FX-1031] feature result pages should have a return to flows button

evanfreeze:
## What
Adds a reset/restart button to the top right of the app bar on all screens that aren't the merchant details or action selection. This will allow testers to jump back to the action selection screen from any subsequent scene regardless of where they are in the flow. It also resets the state to be ready for a new action selection. It persists the merchant details (the assumption being you want to keep the same merchant), but those could easily be changed with another tap of the back arrow after you tap the Reset icon button.

## Why
https://linear.app/joinforage/issue/FX-1031/feature-result-pages-should-have-a-return-to-flows-button

## Test Plan
Manually tested in the android emulator

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/e838679d-d26e-45f1-9b35-b6edd1766b15

## How
Can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
@@ -26,6 +26,7 @@ import com.joinforage.forage.android.ui.ForagePANEditText
 import com.joinforage.forage.android.ui.ForagePINEditText
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.Moshi
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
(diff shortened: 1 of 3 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
```diff
@@ -60,6 +61,22 @@ class POSViewModel : ViewModel() {
         _uiState.update { it.copy(localRefundState = refundState) }
     }
 
+    fun resetUiState() {
+        // this needs to be in a coroutine and delayed to allow for the back-stack
+        // to be fully popped before some of the data it depends on disappears.
+        // There's probably a better way to do this but this works for now.
```
devinmorgan, line 67: Very helpful note!!

devinmorgan (approved):
Cool icon choice! Nice work!

PR #209, opened 2024-02-11
[FX-1051] Fix issue with void receipts having inaccurate data

evanfreeze:
## What
Updates void receipts to have accurate status and balances.

In order to facilitate that^ I had to do a fair amount of refactoring. We were relying on the local tokenizedPaymentMethod and paymentRequest in a few places where they sometimes may not be set. Now that we're clearing the app state with the restart button, these values get cleared when starting a new flow. So, for instance, trying to void something would never have a tokenizedPaymentMethod since you never hit a PAN field in that flow. Before this change, we could've had a subtle bug where the receipt could've shown payment method info that wasn't associated with the actual transaction being voided or refunded.

I went through the receipt flows and figured out what data we needed where and what the right source of truth for it was, then added calls in the view model methods to populate those state fields each time. 

This also adds click-to-copy refs to payment and refund result pages

## Why
https://linear.app/joinforage/issue/FX-1050/feature-add-ref-to-payment-and-refund-receipts
https://linear.app/joinforage/issue/FX-1051/bug-void-receipt-has-inaccurate-data

## Test Plan
Manually tested to verify correct receipt status and balances for both payment and refund voids

## How
Can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
@@ -268,11 +268,17 @@ class POSViewModel : ViewModel() {
 
         viewModelScope.launch {
             try {
-                val response = api.voidPayment(
+                var response = api.voidPayment(
                     idempotencyKey = idempotencyKey,
                     paymentRef = paymentRef
                 )
+                val payment = api.getPayment(paymentRef)
                 val paymentMethod = api.getPaymentMethod(response.paymentMethod)
+                if (response.receipt != null && payment.receipt != null) {
+                    response.receipt!!.isVoided = true
+                    response.receipt!!.balance.snap = (response.receipt!!.balance.snap.toDouble() + payment.receipt!!.snapAmount.toDouble()).toString()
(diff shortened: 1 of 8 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
devinmorgan, line 1: Nice work in this file. I see the `+`s for `Payment`s and `-`s for `Refund`s 👍

sample-app/src/main/java/com/joinforage/android/example/ui/pos/screens/refund/RefundResultScreen.kt
```diff
@@ -39,46 +44,62 @@ fun RefundResultScreen(
...
+                    Row(
+                        verticalAlignment = Alignment.CenterVertically,
+                        horizontalArrangement = Arrangement.Center
+                    ) {
+                        Text("Refund Ref: ${refundResponse.ref}")
+                        Button(onClick = {
+                            clipboardManager.setText(AnnotatedString(refundResponse.ref))
+                        }, colors = ButtonDefaults.elevatedButtonColors()) {
+                            Text("Copy")
+                        }
```
devinmorgan, line 95: BEAUTIFUL!!!!

devinmorgan (approved):
Don't work on weekends! But also, this PR is sick! Great work!

PR #210, opened 2024-02-12
[FX-828] Handle cord-pulled scenarios where we need to refetch payments/refunds on demand to generate receipts

evanfreeze:
## What
Updates the receipt logic to handle the situation where the android SDK doesn't poll for sqs messages (simulating what would happen if the terminal was offline), so the backend auto-voids the payment/refund after it goes unacknowledged by the client for 90 seconds.

This is accomplished by showing a button to refetch the relevant entity instead of displaying a receipt. After the auto-void has occurred in the backend, we can tap the button to refetch the data and display the receipt that accurately reflects the now-voided status

## Why
https://linear.app/joinforage/issue/FX-828/pos-application-needs-to-offer-a-way-to-get-the-last-payment-refund

## Test Plan
- Manually verified the correct behavior in the emulator with the polling disabled in the local android SDK. See demo videos.

## Demo

This is the behavior when we don't poll for a payment...

https://github.com/teamforage/forage-android-sdk/assets/11556475/2b59fdb0-a48f-4a2f-8b48-a0205fd73348

This is the behavior when we don't poll for a refund... 

https://github.com/teamforage/forage-android-sdk/assets/11556475/095fbbdf-f949-460c-ac13-4d89cd5ee154

## How
Can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
@@ -512,11 +512,13 @@ fun POSComposeApp(
                     merchant = uiState.merchant,
                     terminalId = k9SDK.terminalId,
                     paymentMethod = uiState.tokenizedPaymentMethod,
-                    txType = uiState.refundPaymentResponse?.let { it1 -> it1.receipt?.let { it2 ->
-                        TxType.forReceipt(
-                            it2
-                        )
-                    } },
+                    txType = uiState.refundPaymentResponse?.let { it1 ->
+                        it1.receipt?.let { it2 ->
+                            TxType.forReceipt(
+                                it2
+                            )
+                        }
+                    },
                     refundResponse = uiState.refundPaymentResponse!!,
                     fetchedPayment = uiState.capturePaymentResponse,
                     onRefundRefClicked = { paymentRef, refundRef -> viewModel.fetchRefund(paymentRef, refundRef) },
@@ -590,11 +592,13 @@ fun POSComposeApp(
                     merchant = uiState.merchant,
                     terminalId = k9SDK.terminalId,
                     paymentMethod = uiState.tokenizedPaymentMethod,
-                    txType = uiState.voidRefundResponse?.let { it1 -> it1.receipt?.let { it2 ->
-                        TxType.forReceipt(
-                            it2
-                        )
-                    } },
+                    txType = uiState.voidRefundResponse?.let { it1 ->
+                        it1.receipt?.let { it2 ->
+                            TxType.forReceipt(
+                                it2
+                            )
+                        }
+                    },
                     refundResponse = uiState.voidRefundResponse,
                     onBackButtonClicked = { navController.popBackStack(POSScreen.VOIDRefundScreen.name, inclusive = false) },
                     onDoneButtonClicked = { navController.popBackStack(POSScreen.ActionSelectionScreen.name, inclusive = false) }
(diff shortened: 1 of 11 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/screens/refund/RefundResultScreen.kt
```diff
@@ -45,7 +49,20 @@ fun RefundResultScreen(
...
+                    Button(onClick = onReloadButtonClicked) {
+                        Text("Re-fetch Payment")
+                    }
+                } else {
+                    Text("Select a Refund ref from payment (${fetchedPayment.ref}) to view the receipt for:")
+                    fetchedPayment.refunds.forEach { refundRef ->
+                        Button(onClick = { onRefundRefClicked(fetchedPayment.ref!!, refundRef) }) {
+                            Text(refundRef)
+                        }
+                    }
```
devinmorgan, line 64: I cannot believe that making a list of `Button`s is as easy as doing a `forEach` loop 🤯🤯🤯🤯. Compose is soooo effing cool!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
```diff
@@ -507,10 +512,25 @@ fun POSComposeApp(
...
                     terminalId = k9SDK.terminalId,
                     paymentMethod = uiState.tokenizedPaymentMethod,
-                    txType = uiState.refundPaymentResponse?.let { it1 -> TxType.forReceipt(it1.receipt) },
+                    txType = uiState.refundPaymentResponse?.let { it1 ->
+                        it1.receipt?.let { it2 ->
+                            TxType.forReceipt(
+                                it2
+                            )
+                        }
+                    },
```
devinmorgan, line 521: haha `it1` and `it2`. Are these variables basically `refund` and `receipt` respectively?

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
```diff
@@ -507,10 +512,25 @@ fun POSComposeApp(
...
                     refundResponse = uiState.refundPaymentResponse!!,
+                    fetchedPayment = uiState.capturePaymentResponse,
+                    onRefundRefClicked = { paymentRef, refundRef -> viewModel.fetchRefund(paymentRef, refundRef) },
```
devinmorgan, line 524: I'm still a little bit bitter that the `Refund` endpoint requires the `paymentRef` in the URL 🙄

devinmorgan (approved):
Cannot believe that you shipped `Refund`s in this too! Amazing!!

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
evanfreeze: yeah, exactly. This was an automated refactor thing in android studio and I just went with it 😅

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
evanfreeze: Yeahhh maybe not ideal

sample-app/src/main/java/com/joinforage/android/example/ui/pos/screens/refund/RefundResultScreen.kt
evanfreeze: Right?! I had to look it up and was like "oh wow that's super easy"

PR #212, opened 2024-02-13
[FX-1019] Add support for printing receipts for payment declines

evanfreeze:
## What
Adds support for printing receipts for declined payments. We're essentially cobbling together a manual receipt since we don't really get any receipt data back from the backend when a payment is declined. There is probably be a better way to do this, but this works?

## Why
https://linear.app/joinforage/issue/FX-1019/error-handling-need-a-receipt-scene-for-failed-payment-captures

## Test Plan
- Manually tested both insufficient funds and incorrect/invalid PIN and saw declined receipts

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/596c8011-df98-4882-b41c-8b181f7eb257

## How
Can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/SnapPurchaseTxReceipt.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/SnapPurchaseTxReceipt.kt
@@ -1,7 +1,7 @@
 package com.joinforage.android.example.pos.receipts.templates.txs
 
 import com.joinforage.android.example.ui.pos.data.Merchant
-import com.joinforage.android.example.ui.pos.data.PosPaymentResponse
+import com.joinforage.android.example.ui.pos.data.Receipt
 import com.joinforage.android.example.ui.pos.data.Refund
 import com.joinforage.android.example.ui.pos.data.tokenize.PosPaymentMethod
 
@@ -11,8 +11,8 @@ internal class SnapPurchaseTxReceipt : TxReceiptTemplate {
         merchant: Merchant?,
         terminalId: String,
         paymentMethod: PosPaymentMethod?,
-        payment: PosPaymentResponse
-    ) : super(merchant, terminalId, paymentMethod, payment.receipt!!)
+        receipt: Receipt
+    ) : super(merchant, terminalId, paymentMethod, receipt)
 
     constructor(
         merchant: Merchant?,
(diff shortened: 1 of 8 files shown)


sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/SnapPurchaseTxReceipt.kt
```diff
@@ -11,8 +11,8 @@ internal class SnapPurchaseTxReceipt : TxReceiptTemplate {
         merchant: Merchant?,
         terminalId: String,
         paymentMethod: PosPaymentMethod?,
-        payment: PosPaymentResponse
-    ) : super(merchant, terminalId, paymentMethod, payment.receipt!!)
+        receipt: Receipt
+    ) : super(merchant, terminalId, paymentMethod, receipt)
```
devinmorgan, line 15: Given that the constructor accepts a `Receipt` instead of the `PosPaymentResponse`, is the second constructor even needed anymore? I only made two constructors before so that it was easy to handle both `Payment`s and `Response`s but if we're just passing the receipt directly, it's unclear whether the second constructor has a point anymore 🤷

sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSComposeApp.kt
```diff
@@ -447,16 +447,39 @@ fun POSComposeApp(
...
+                    balance = ReceiptBalance(
+                        id = 0.toDouble(),
+                        snap = uiState.tokenizedPaymentMethod!!.balance!!.snap,
+                        nonSnap = uiState.tokenizedPaymentMethod!!.balance!!.non_snap,
+                        updated = uiState.tokenizedPaymentMethod!!.balance!!.updated
+                    ),
+                    created = uiState.createPaymentResponse!!.created,
+                    message = uiState.capturePaymentError ?: "Unknown Error"
+                )
+
```
devinmorgan, line 471: OMG it sounds like you *definitely* went on a journey yesterday 😅

devinmorgan (approved):
This is great! 

NOTE TO SELF: One thing that I think we'll want to pay attention to during the Wet Rehearsal today is whether the messages that are getting printed on the receipts are OK. Right now the client dumps the `receipt.message` and we should probably just sanity check that the messages are OK as is

sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/SnapPurchaseTxReceipt.kt
evanfreeze: Yeah I thought about this too. Seems like we might be able to reduce it down to just one if we stick with this approach (imo the sdk/api should return a receipt in an error scenario too so this whole thing becomes unnecessary). 

I liked having a constructor for each, but for these decline receipts it was a lot easier to only need to create a manual receipt instead of having to create an entire payment or refund.

PR #214, opened 2024-02-13
[FX-1047] Fix issue where error messages from previous entries would persist

evanfreeze:
## What
Clears PAN & PIN errors when navigating forward to those screens to avoid showing potentially stale error messages

## Why
https://linear.app/joinforage/issue/FX-1047/bug-error-messages-from-previous-entries-persist-across-navigation

## Test Plan
- Manually verified in emulator (see demo)

## Demo
https://github.com/teamforage/forage-android-sdk/assets/11556475/45c7a44b-6cb2-4d17-bf2e-b42c4c2e1c9c

## How
Can be merged as-is


devinmorgan (approved):
Nice fixes!!

PR #216, opened 2024-02-14
[FX-1086, FX-1085, FX-1046] Improve declined receipts, fix terminal ID, avoid unknown sequence numbers

evanfreeze:
## What
- Improves declined receipts by refetching the payment when a decline fails, which allows us to read the `receipt` field and use that data to populate the declined receipt instead of my cobbled together version.
  - Also tidies up the error message to only show the message response instead of the full stringified error
- Updates the terminal ID to read from the receipt's `posTerminal.terminalId` so we're showing the forage ref value
- Updates the logic for PIN errors to only route to a receipt if we have a sequence number, which means we actually tried to capture the payment. 
  - For invalid PIN errors like only typing in 3 digits, we never make it to FIS so there's no sequence number, so we won't show a receipt. Instead we'll just show an inline error message. If we _do_ have a sequence number, like in the case of entering an incorrect 4-digit PIN, we'll still show a receipt since we would've actually tried to capture the payment at that point. See [this Slack discussion](https://joinforage.slack.com/archives/C061FR1EW2Y/p1707943851160979) for context.

## Why
https://linear.app/joinforage/issue/FX-1086/declined-error-messages-are-noisy
https://linear.app/joinforage/issue/FX-1085/unknown-sequence-number-on-declined-receipts
https://linear.app/joinforage/issue/FX-1046/terminal-id-on-receipt-is-the-from-the-cpay-sdk

## Test Plan
Manually verified in the emulator

## How
Can be merged as-is

### DIFF
--- a/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
+++ b/sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
@@ -305,7 +305,6 @@ class POSViewModel : ViewModel() {
                     val jsonAdapter: JsonAdapter<Refund> = RefundJsonAdapter(moshi)
                     val refundResponse = jsonAdapter.fromJson(response.data)
                     _uiState.update { it.copy(refundPaymentResponse = refundResponse, refundPaymentError = null) }
-                    Log.i("POSViewModel-REFUND", refundResponse.toString())
                     onSuccess()
                 }
                 is ForageApiResponse.Failure -> {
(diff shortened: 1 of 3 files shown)


sample-app/src/main/java/com/joinforage/android/example/ui/pos/POSViewModel.kt
```diff
@@ -303,7 +309,18 @@ class POSViewModel : ViewModel() {
                 }
                 is ForageApiResponse.Failure -> {
                     Log.e("POSViewModel", response.toString())
-                    _uiState.update { it.copy(refundPaymentError = response.errors.joinToString("\n"), refundPaymentResponse = null) }
+                    var payment: PosPaymentResponse? = null
+                    var refund: Refund? = null
+                    try {
+                        payment = api.getPayment(paymentRef)
+                        val mostRecentRefundRef = payment.refunds.lastOrNull()
```
devinmorgan, line 316: great variable name 👌

devinmorgan (approved):
this one looked like a precision surgery - nice work!

PR #217, opened 2024-02-15
[FX-1066] Handle late-notification scenarios by showing voided receipts

evanfreeze:
## What
Handles the late-notification payment scenarios by showing voided receipts (I think this is what we want? it's easy to make it a decline if that's what we want instead)

## Why
https://linear.app/joinforage/issue/FX-1066/handle-late-notification-responses-from-payments-api

## Test Plan
Manually verified in emulator

## How
Can be merged as-is


evanfreeze:
CI test failure isn't related to these changes and is being addressed here: https://joinforage.slack.com/archives/C057H6RGPU0/p1708012798977519

devinmorgan (approved):

PR #218, opened 2024-02-15
[FX-1083, FX-1084] Fix receipt discrepancies

evanfreeze:
## What
Fixes outstanding issues with payment and refund receipts

## Why
https://linear.app/joinforage/issue/FX-1083/payment-receipt-inaccuracies
https://linear.app/joinforage/issue/FX-1084/refund-receipt-inaccuracies

## Test Plan
Manually verified all receipt flows in the emulator and they're all working as expected 🥳

## How
Can be merged as-is


sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/CashWithdrawalTxReceipt.kt
```diff
@@ -10,12 +10,13 @@ internal class CashWithdrawalTxReceipt : TxReceiptTemplate {
...
-    ) : super(merchant, terminalId, paymentMethod, receipt)
+        receipt: Receipt,
+        title: String
+    ) : super(merchant, terminalId, paymentMethod, receipt, title)
 
     override val txContent = CashWithdrawalLayout(
         snapBal,
         cashBal,
-        receipt.cashBackAmount
+        receipt.ebtCashAmount
```
evanfreeze, line 20: A withdrawal is indicated by a combo of `funding_type: "ebt_cash"` + `transaction_type: "withdrawal"` as best as I can tell from what is in [this linear issue](https://linear.app/joinforage/issue/FX-748/story-cash-withdrawal-flow), so we want to use the ebtCashAmount here.

sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/TxDataLayout.kt
```diff
@@ -55,5 +55,5 @@ internal class CashWithdrawalLayout(
     override val withdrawalAmt: String
 ) : TxDataLayout() {
     override val snapAmt: String = ZERO_TX
-    override val cashAmt: String = ZERO_TX
+    override val cashAmt: String = withdrawalAmt
```
evanfreeze, line 58: This sample receipt from the certification script seems to suggest the `CASH` line should always have a value if the cash was touched at all. So if it was a withdrawal for $2.00, we should see `CASH 2.00` and `CS W/D 2.00` based on my understanding.

In the cash-back scenario where there's both a purchase and cash back, the `CASH` value would be the combination of the purchase and the cashback. So a $2.00 purchase with $2.00 cashback would be `CASH 4.00` and `CS W/D 2.00`, again based on my interpretation of this / understanding.

sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/TxType.kt
```diff
@@ -60,37 +58,19 @@ enum class TxType(val title: String) {
...
             }
-            if (spentEbtCash(payment) && !withdrewEbtCash(payment)) {
+            if (fundingType == FundingType.EBTCash.value) {
                 return CASH_PAYMENT
             }
-
-            return UNKNOWN
-        }
-
-        fun forRefund(payment: PosPaymentRequest): TxType {
```
evanfreeze, line 80: We weren't using this forRefund one anyway – we're using the forReceipt one for the refund screen right now

sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/CashWithdrawalTxReceipt.kt
devinmorgan: I think you're right on this one. We can confirm in the wet rehearsal tomorrow when all eyes are there but I vaguely remember that `cashBackAmount` did not always refer to the `withdrawal` amount or the `cashback` amount, which is what I mistakenly original thought the `cashBackAmount` meant. Nice catch!

sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/TxDataLayout.kt
devinmorgan: yup I agree with this. This definitely was confusing me so I'm glad you were able to make sense of it for the rest of us 🙌

sample-app/src/main/java/com/joinforage/android/example/pos/receipts/templates/txs/TxType.kt
```diff
@@ -60,37 +58,35 @@ enum class TxType(val title: String) {
...
+            if (fundingType == FundingType.EBTSnap.value) {
+                return SNAP_PAYMENT
             }
-            if (spentEbtCash(payment) && !withdrewEbtCash(payment)) {
+            if (fundingType == FundingType.EBTCash.value) {
                 return CASH_PAYMENT
             }
-
             return UNKNOWN
         }
```
devinmorgan, line 75: Wow I feel like this is the code expression of the notebook you showed us😵‍💫. Thanks for getting to the bottom of this!

devinmorgan (approved):
Nice! You crushed it this week. Thanks Evan!

PR #226, opened 2024-02-29
[TW-406] Update comments for states, derivedCardInfo, USState

kimberleehowley:
## What

This PR updates comments for states, derivedCardInfo, and USState. 

## Why

These comments will eventually serve as customer-facing docs. 

## Test Plan

- ❌ | Tests aren't needed for comment changes. 
- ✅ | The reviewer should manually test the changes in this PR. 

Check out my branch, and then: 

```
./gradlew dokkaHtml
npx http-serve reference-docs
```

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
@@ -63,7 +63,7 @@ internal data class DerivedCardInfoDto(
 /**
  * An interface that represents the state of a
  * [ForagePANEditText][com.joinforage.forage.android.ui.ForagePANEditText] Element.
- * @property derivedCardInfo The [derivedCardInfo] for the card number submitted to the Element.
+ * @property derivedCardInfo The [DerivedCardInfo] for the card number submitted to the Element.
  */
 interface PanElementState : ElementState {
     val derivedCardInfo: DerivedCardInfo // the interface not the DTO
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
@@ -8,10 +8,12 @@ import com.joinforage.forage.android.model.USState
  * as the customer interacts with it.
  * @property isFocused Whether the Element is in focus.
  * @property isBlurred Whether the Element is blurred.
- * @property isEmpty Whether the text field of the Element is empty.
- * @property isValid Whether the text field contains valid text.
- * @property isComplete Whether the text field of the Element is ready to submit.
- * @property validationError A [ElementValidationError], if applicable.
+ * @property isEmpty Whether the input value of the Element is empty.
+ * @property isValid Whether the input text fails any validation checks, with the exception of the
+ * target length constraint.
+ * @property isComplete Whether the text field of the Element is ready to submit. This is `true`
+ * when all input value validation constraints are satisfied.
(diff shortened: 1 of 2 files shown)


forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
```diff
@@ -32,6 +46,10 @@ internal val INITIAL_PIN_ELEMENT_STATE = PinElementStateDto(
     validationError = null
 )
 
+/**
+ * An interface that represents information that Forage gets from the card.
+ * This includes the [USState] that issued the card.
```
djoksimo, line 51: lovin the use of links/refs in this PR!

forage-android/src/main/java/com/joinforage/forage/android/model/StateIIN.kt
```diff
@@ -2,6 +2,10 @@ package com.joinforage.forage.android.model
 
 internal const val STATE_INN_LENGTH = 6
 
+/**
+ * A class that includes all of the possible US states that issue EBT Cards and their corresponding
+ * abbreviations.
```
djoksimo, line 7: Can we add a remark that South & North Dakota-issued cards have the same IIN?

@dleis612 to correct me in case I missed something.

djoksimo (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
```diff
@@ -12,6 +22,10 @@ interface ElementState {
     val validationError: ElementValidationError?
 }
 
+/**
+ * An interface that represents the state of a
```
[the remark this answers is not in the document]
kimberleehowley: @djoksimo I'm open to this! Does this look like what you had in mind? 

```
{
  "isFocused": true,
  "isBlurred": false,
  "isEmpty": false,
  "isValid": true,
  "isComplete": true,
  "validationError": null
}
```

If not I'd appreciate it if you could share your ideal example 🙏  I asked ChatGPT in case I was missing something.

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
```diff
@@ -40,6 +60,11 @@ internal data class DerivedCardInfoDto(
     override val usState: USState? = null
 ) : DerivedCardInfo
 
+/**
+ * An interface that represents the state of a
+ * [ForagePANEditText][com.joinforage.forage.android.ui.ForagePANEditText] Element.
+ * @property derivedCardInfo The [derivedCardInfo] for the card number submitted to the Element.
```
djoksimo, line 66:
```suggestion
 * @property derivedCardInfo The [DerivedCardInfo] for the card number submitted to the Element.
```

forage-android/src/main/java/com/joinforage/forage/android/core/element/state/ElementState.kt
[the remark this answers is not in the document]
djoksimo: Sry, I just meant to mention a brief summary of what `PinElement` state includes: 

> An interface that represents the state of a .... . This includes input validation values....

But given that you link to `ElementState` interface, I don't think this is necessary! So no action needed here

djoksimo (changes requested):
Left a small comment re: derivedCardInfo, but everything else LGTM

djoksimo (approved):

PR #227, opened 2024-03-01
[FX-1128] Remove get merchants call from pos cert app

evanfreeze:
## What
Removes the call to get merchant details from the POS cert app and updates the UI to use a hard-coded test merchant for the merchant name and address. Merchant ID will still read from whatever is provided in the text field in Merchant Setup.

## Why
https://linear.app/joinforage/issue/FX-1128/remove-get-merchants-call-from-pos-app

## Test Plan
- Manually verified in Android Emulator that expected flows still succeed with mock merchant data using a non-internal token

## How
Can be merged as-is


djoksimo (changes requested):
Thoughts on hiding the Merchant Name and Address from this page?

I think the discrepancy could lead merchants to confusion

image

djoksimo (approved):

PR #233, opened 2024-03-16
Move PosDeferPaymentRefundParams to the right module

djoksimo:
- Move `PosDeferPaymentRefundParams` to the right directory. 

- This ensures POS docs don't pollute the online docs
- Co-locate POS-related imports

⚠️ Technically a breaking change. [Learn more](https://linear.app/joinforage/issue/FX-1161/move-posdeferpaymentrefundparams-to-the-pos-directorymodule)

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKInterface.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ForageSDKInterface.kt
@@ -166,22 +166,3 @@ data class DeferPaymentCaptureParams(
     val foragePinEditText: ForagePINEditText,
     val paymentRef: String
 )
-
-/**
- * A model that represents the parameters that Forage requires to collect a card PIN and defer
- * the refund of the payment to the server.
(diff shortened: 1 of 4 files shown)


devinmorgan (approved):
This is waaaaay too small of a change to be warrant calling for a new major version. I see two realities:
1. Our partners will not experience this change - We originally told our partner to use the `sonatype` hosted regular SDK so at some point, this would have definitely been felt by our partner. However, I understand that they now rely on the private repo, which is *not* the public repo. If that's true, our partner will not experience the effect of this
2. Our partner does experience the effect of this - they can deal with a minor change like this provided we communicate this to them. **However, we can't do this all the time.** Fortunately, the next time our partner is going to try our SDK is April 18th. If we were going to give them a new release every week, I'd feel a bit more uneasy about this. So, I think it's fine to make this change provided [we keep track of this to remind our partner about it](https://www.notion.so/joinforage/Release-Tracker-94d996c2c7e14501a96d54722a89532e?pvs=4#831ded82152c423ab63e33964eb0546d)

PR #234, opened 2024-03-16
Fixes to facilitate keeping private + public repos in sync

djoksimo:
- Changes that should be applied in both this repo and the private fork one.
- The private repo uses more than just `/proxy/` endpoints.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/vault/ForagePinSubmitter.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/vault/ForagePinSubmitter.kt
@@ -109,7 +109,7 @@ internal class ForagePinSubmitter(
         private fun buildVaultUrl(path: String): HttpUrl =
             VAULT_BASE_URL.toHttpUrlOrNull()!!
                 .newBuilder()
-                .scheme("https")
+                .addPathSegment("proxy")
                 .addPathSegmentsSafe(path)
                 .addTrailingSlash()
                 .build()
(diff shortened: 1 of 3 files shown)


forage-android/src/main/java/com/joinforage/forage/android/vault/ForagePinSubmitter.kt
```diff
@@ -109,7 +109,7 @@ internal class ForagePinSubmitter(
         private fun buildVaultUrl(path: String): HttpUrl =
             VAULT_BASE_URL.toHttpUrlOrNull()!!
                 .newBuilder()
-                .scheme("https")
```
djoksimo, line 112: turns out this wasn't necessary after all and impeded local development against http.

devinmorgan: Nice!

devinmorgan (commented):
So I think there is a bit of a design question going on here around what URL information should be centralized in `EnvConfig` vs can live inside of the `buildVaultUrl()` method.

The proposed change to move `proxy/` to `.addPathSegment()` would make a lot of sense if we anticipate or already use multiple paths for the `vault.joinforage.app` host. For example:
* `vault.joinforage.app/some_path` 
* `vault.joinforage.app/another_path` 

In that world, the "common part" to would be `vault.joinforage.app` and it would make sense in my mind that specific `buildVaultUrl()`s would customize the `/path` portion for their own needs (e.g. adding `/some_path` or `/another_path`)

However, if all current and foreseeable usages of `vault.joinforage.app` are always `vault.joinforage.app/path`, then I think we're decentralizing information a portion of the "common part", which (albeit slightly) hurts readability (now you need to go to two files to understand that we're hitting `vault.joinforage.app/path` instead of just one file `EnvConfig`).

So, my two cents are:
1. 🆗 deleting of the `.scheme("https")`
2. 🚫 let's no include the changes that add `.addPathSegment("proxy")`, unless we use multiple paths on `vault.joinforage.app`

devinmorgan (approved):
Nice!

PR #236, opened 2024-03-25
Introduce init method in the public repo

djoksimo:
## What

- [x] Introduced an empty `ForageTerminalSDK.init` method
- [x] Removed outdated constructor info and made the constructor internal
- [x] `init` method throws a `NotImplementedError` if the client calls the `init` method from the wrong SDK distribution, unless an override ID is provided. 
- [x] Corrected sample app to use `init` method 
- [x] Added updated comments for `init` method

## Why 

ℹ️ The method's implementation is not available publicly, but these changes facilitate keeping the different repositories in sync and **build our public-facing documentation**

## Test plan

- [x] Manually ran some POS operations in the sample app
- [x] Unit tests pass
- [x] Unit tests were updated to be in sync with the private repo 
- [x] Built docs locally via `./gradlew dokkaHtml` task

PR #237, opened 2024-03-29
[FX-1225] Support defer capture + defer refund in POS sample app

djoksimo:
## What

- See title

## Why/How

- Deferred actions are the primary (primary Material Design color + shape + position) CTAs, since deferring actions to server is the preferred flow
- Based on #236 since PR 236 includes changes for invoking the `init` method from the sample app, to facilitate keeping the private POS repo in sync

## Test plan

- Manually tried some captures + refunds in the sample app
- Ensured existing non-deferred POS flows still work :)


devinmorgan (approved):
Nice work! Does this PR make all of the changes necessary for running CI tests in the Sample App that we are currently aware of (so excluding unknowns) or does this PR prepare of the Bug Bash and still leave some outstanding work for CI Tests?

djoksimo:
> Nice work! Does this PR make all of the changes necessary for running CI tests in the Sample App that we are currently aware of (so excluding unknowns) or does this PR prepare of the Bug Bash and still leave some outstanding work for CI Tests?

@devinmorgan Great question ; this PR (#237) prepares us for the Bug Bash

This [small ticket](https://linear.app/joinforage/issue/FX-1235/add-test-ids-to-pos-sample-app) for adding test IDs is part of the outstanding work for CI tests.

PR #238, opened 2024-04-02
[FX-1184] Add the default header to lock all endpoints

dleis612:
## What
Add the default version header to all endpoints.

## Test Plan
Will test this locally against BT and VGS.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/network/OkHttpClientBuilder.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/network/OkHttpClientBuilder.kt
@@ -20,9 +20,11 @@ internal object OkHttpClientBuilder {
                             "${ForageConstants.Headers.BEARER} $sessionToken"
                         )
                         .run {
+                            // If the API_VERSION header has already been appended, don't override it!
                             chain.request().headers[ForageConstants.Headers.API_VERSION]?.let {
                                 this
                             }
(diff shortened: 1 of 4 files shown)


forage-android/src/main/java/com/joinforage/forage/android/network/OkHttpClientBuilder.kt
```diff
@@ -18,7 +18,17 @@ internal object OkHttpClientBuilder {
...
+                        .run {
+                            chain.request().headers[ForageConstants.Headers.API_VERSION]?.let {
+                                this
+                            }
+                                ?: addHeader(
+                                    ForageConstants.Headers.API_VERSION,
+                                    "default"
+                                )
+                        }
+                        .run {
```
djoksimo, line 31: can you add a comment to explain what's going on in this block pls

djoksimo (approved):

forage-android/src/main/java/com/joinforage/forage/android/network/OkHttpClientBuilder.kt
dleis612: Done: [19bcdaf](https://github.com/teamforage/forage-android-sdk/pull/238/commits/19bcdafe2bebf8b9e8b7ab1bff69dbf4e58cc41a) also tested all flows locally.

PR #249, opened 2024-05-04
[FX-1332] Typed SDK response helpers

djoksimo:
## What

Introduce: `toPaymentMethod()`, `toBalance`, `toPayment`

- Introduce types/helpers for casting SDK responses (JSON string -> PaymentMethod, Balance, Payment, ...) ; non-breaking!
- Added Kdoc comments
- Update existing internal code + tests to be compatible
- [x] Add more unit tests

The new `toPaymentMethod`, `toBalance`, `toPayment` are **primarily intended for e-commerce use**, and similar methods can be introduced for POS-specific types and differences (refunds, ...)

## Why

- Used sealed classes for Balance + Card, like we do for ebt_error_51, to make the interfaces extensible to other payment method types
- Used Parcelable to help serialize responses ; seems conventional

- [Technical Design Document](https://www.notion.so/joinforage/Typed-Android-SDK-Responses-298fa77d1ef64526ae5477581807147e)

## Test Plan

- ✅ Updated unit tests
- ✅ Call `toPaymentMethod`, `toBalance`, `toPayment`

## How

MINOR-level version change


forage-android/src/main/java/com/joinforage/forage/android/model/Card.kt
```diff
@@ -1,17 +1,13 @@
...
-import kotlinx.parcelize.Parcelize
-
 /**
  * @property last4 The last 4 digits of the card number
  */
-@Parcelize
-sealed class Card(open val last4: String, internal open val token: String) : ForageModel {
+sealed class Card(open val last4: String, internal open val token: String) {
     /**
      * @property last4 The last 4 digits of the EBT card number
```
kimberleehowley, line 8: Super nit: The previous descriptions of `Balance` params include periods, but these property descriptions do not. Could we pick one for consistency? It looks like the Android docs tend to use periods ([example](https://android.joinforage.app/forage-android/com.joinforage.forage.android/-check-balance-params/)).

kimberleehowley: I also think we want to capitalize EBT card -> EBT Card, even though we're not referring to the card as a standalone entity.

kimberleehowley (approved):
Great work, thank you so much! I left one small super nit comment.

forage-android/src/main/java/com/joinforage/forage/android/model/Card.kt
djoksimo: Great catches! Addressed here https://github.com/teamforage/forage-android-sdk/pull/249/commits/20466f363a30f85a3a9ef246c720b1ae11f1ce4e

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -98,10 +98,16 @@ class ForageSDK : ForageSDKInterface {
      *
      *         when (response) {
      *             is ForageApiResponse.Success -> {
-     *                 // parse response.data for the PaymentMethod object
+     *                 val paymentMethod = response.toPaymentMethod()
+     *                 // Unpack paymentMethod.ref, paymentMethod.card, etc.
+     *                 val card = paymentMethod.card
+     *                 // Unpack card.last4, ...
+     *                 val ebtCard = card as EbtCard
```
devinmorgan, line 105: *I'm not trying to propose any new designs here in this comment. I'm only trying to surface that the lack of reflection in this example snippet seems different from our discussion*

If we can cast `paymentMethod.card` as an `EbtCard` without doing any `if-checking`,  that means that `paymentMethod.card` is always an `EbtCard`. This behavior does not seem to align with our discussion yesterday. My understanding was that Forage would guarantee that `.card` would satisfy an `interface Card` so the merchant could access properties common to all cards for as long as possible. 

And if the merchant runs into a case where they absolutely must work with more specific information, their options would be:
1. do reflection to inspect `if (paymentMethod.card is EbtCard) { card as EbtCard }` or
2. Forage would expose something like `paymentMethod.card.ebt` to offer the more specific fields or
3. Forage would return something that is type specific (e.g. PaymentMethod<EbtCard>) 
4. Something else?

forage-android/src/main/java/com/joinforage/forage/android/ForageSDK.kt
```diff
@@ -167,10 +173,13 @@ class ForageSDK : ForageSDKInterface {
      *
      *         when (response) {
      *             is ForageApiResponse.Success -> {
-     *                 // response.data will have a .snap and a .cash value
+     *                 val balance = response.toBalance()
+     *                 val ebtBalance = balance as EbtBalance
```
devinmorgan, line 177: Similar question here for `balance as EbtBalance` as I have in [this comment](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/249/FX-1332-Typed-SDK-response-helpers?panel=files#comment-PRRC_kwDOIQqcBs5e60mt).

forage-android/src/main/java/com/joinforage/forage/android/network/data/CapturePaymentRepository.kt
```diff
@@ -33,11 +33,11 @@ internal class CapturePaymentRepository(
             else -> return response
         }
         val payment = when (val response = paymentService.getPayment(paymentRef)) {
-            is ForageApiResponse.Success -> Payment.ModelMapper.from(response.data)
+            is ForageApiResponse.Success -> Payment(response.data)
```
devinmorgan, line 36: omg this looks so clean 🤤

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
```diff
@@ -0,0 +1,40 @@
...
+        internal fun fromSdkResponse(jsonString: String): EbtBalance {
+            val jsonObject = JSONObject(jsonString)
+            val snap = jsonObject.getString("snap")
+            val cash = jsonObject.getString("cash")
+            return EbtBalance(
+                snap = snap,
+                cash = cash
+            )
+        }
+    }
```
devinmorgan, line 35:
```suggestion
constructor(jsonString: String) : this(JSONObject(jsonString)
```
Thoughts on getting rid of the companion object entirely with another constructor `jsonString: String`?

forage-android/src/main/java/com/joinforage/forage/android/network/model/Card.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.joinforage.forage.android.getStringOrNull
+import com.joinforage.forage.android.model.USState
+import org.json.JSONObject
+
+/**
+ * @property last4 The last 4 digits of the card number.
+ */
+interface Card {
+    val last4: String
+}
```
devinmorgan, line 12: Nice! Excited for this!

forage-android/src/main/java/com/joinforage/forage/android/network/model/ForageApiResponse.kt
```diff
@@ -21,8 +21,69 @@ sealed class ForageApiResponse<out T> {
...
+         *     }
+         * }
+         * ```
+         * @return A [Balance] instance.
+         */
+        fun toBalance(): Balance {
+            // The ForageApiResponse.data string is already formatted to
+            // { snap: ..., cash: ... }
+            // so we use fromSdkResponse() instead of the typical constructor
+            return EbtBalance.fromSdkResponse(data as String)
```
devinmorgan, line 68:
```suggestion
            return EbtBalance(data as String)
```
If we take [this change](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/249/FX-1332-Typed-SDK-response-helpers?panel=files#comment-PRRC_kwDOIQqcBs5e7Jyx), then this line could be simplified a bit

forage-android/src/main/java/com/joinforage/forage/android/network/model/Payment.kt
```diff
@@ -0,0 +1,147 @@
...
+    val metadata: Map<String, String>?,
+    val paymentMethodRef: String,
+    val receipt: Receipt?,
+    val ref: String,
+    val refunds: List<String>,
+    val status: String,
+    val successDate: String?,
+    val updated: String
+) {
+    internal constructor(jsonString: String) : this(JSONObject(jsonString)) // you get json string constructor for free!
```
devinmorgan, line 116: haha feel free to delete this comment!

forage-android/src/main/java/com/joinforage/forage/android/network/model/Payment.kt
```diff
@@ -0,0 +1,147 @@
...
+ * @property paymentMethodRef The unique reference hash for the existing Forage PaymentMethod that was charged in this transaction.
+ * @property receipt Most of the information that you're required to display to the customer, according to FNS regulations.
+ * [receipt] is null if the data that populates the receipt is not yet available.
+ * @property ref A string identifier that refers to an instance in Forage's database of a Payment object,
+ * which is a  one-time charge to a previously created PaymentMethod.
+ * @property refunds A list of unique reference hashes for the Refund objects
+ * that were created for this Payment.
+ * @property status The status of the Payment. [Learn more](https://docs.joinforage.app/reference/payments#payment-lifecycle)
+ * @property successDate A UTC-8 timestamp of when the Payment was successfully processed, represented as an ISO 8601 date-time string.
+ * @property updated A UTC-8 timestamp of when the Payment was last updated, represented as an ISO 8601 date-time string.
```
devinmorgan, line 97: Thanks for putting in the effort for such clear docstrings!

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
```diff
@@ -0,0 +1,40 @@
+package com.joinforage.forage.android.network.model
+
+import org.json.JSONObject
+
+/**
+ * A model that represents the available balance on a customer’s card.
+ * @see [EbtBalance]
+ */
+interface Balance
```
devinmorgan, line 9: I'm totally aligned on the value of putting out the `interface Card` right now along side the `class EbtCard` because I agree that every card will have a `last4`.

I do wonder if the same pattern is necessary for balances if, out of the gate, we can't identify any commonalities that the `EbtBalance` would share with another balance. The need to add an empty interface to our public API feels a bit preemptive and it's something we can't take back. Thoughts on not rolling out the `interface Balance` right now and only rolling it out later if we ever identify something common across all balances?

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
djoksimo: @devinmorgan 

```kotlin
// the JSON object constructor is not compatible with the "SDK response"
 constructor(jsonObject: JSONObject) : this(
        jsonObject.getString("snap"),
        jsonObject.getString("non_snap")
    )
```

The other (`Payment`, `PaymentMethod`, ...) constructors map the raw/unprocessed JSON response from the Payments API.

Whereas the Balance's `non_snap` field is already mapped/renamed to `cash`, which is why I need a custom `fromSdkResponse` constructor – to make it explicitly clear that the factory is only compatible with the "SDK response" as input.

Alternatively, I can make a constructor like `cash = json.get('non_snap') ?: json.get('cash')` – that is compatible with either format.

What are your thoughts on the alternative?

djoksimo: In any case, this code will likely get simplified when we introduce _actual_ typed SDK responses (instead of JSON strings)

devinmorgan (commented):
Thanks for all your effort in this PR, Danilo. My remaining questions are:
1. were the `.idea/` files intentionally included in the PR?
2. Is it necessary to include the empty `interface Balance` right now?
3. can we drop `EbtBalance.fromSdkResponse(data as String)` with another constructor?

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
djoksimo: @devinmorgan 

> only rolling it out later if we ever identify something common across all balances

- We want clients to design/build around generic-named types/interfaces.
- If we don't support this interface from the beginning, we have to:
  - now: a) make the return type of `toBalance() -> EbtBalance // non-generic name :(`
  - in the future: b) introduce the `interface Balance` to support `toBalance() -> Balance // breaking change :(` 

> we can't identify any commonalities that the

I see a world where `updatedAt / updated_at` is a common field between different kinds of balances (like it is in the [JS SDK](https://github.com/teamforage/fe-monorepo/blob/79e6dade8cd21c83f43ce14dd7e96b4ca0371281/apps/sdk/src/common/interfaces/Forage/types.ts#L344)). I didn't expose it in this PR, because the previous SDK version did not expose `updated_at`, and we can introduce it when it is requested / future PR.

djoksimo:
> Thanks for all your effort in this PR, Danilo. My remaining questions are:
> 
> 1. were the `.idea/` files intentionally included in the PR?
> 2. Is it necessary to include the empty `interface Balance` right now?
> 3. can we drop `EbtBalance.fromSdkResponse(data as String)` with another constructor?

@devinmorgan 

1) bb596f3665535f306e5c82ff39cc8f7b73bc0c6b nice catch ; accident

2) replied: https://github.com/teamforage/forage-android-sdk/pull/249#discussion_r1592622026

3) replied: https://github.com/teamforage/forage-android-sdk/pull/249#discussion_r1592594254

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
devinmorgan: > * If we don't support this interface from the beginning, we have to:
>   
>   * now: a) make the return type of `toBalance() -> EbtBalance // non-generic name :(`
>   * in the future: b) introduce the `interface Balance` to support `toBalance() -> Balance // breaking change :(`

Ah OK good point about needing to change the method signature of `toBalance()` if we don't release it now. I'm convinced!

One follow up ask is if can we update `FlowBalanceViewModel.kt` to use reflection like `if (balance is EbtBalance) { balance as EbtBalance }` instead of just straightaway casting `balance as EbtBalance`

This way the Sample App will also document the conventions that we're advocating merchants follow so we don't forget internal. That OK with you?

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
devinmorgan: cool sounds like it's no a quick win so let's skip for now! Thanks for considering!

forage-android/src/main/java/com/joinforage/forage/android/network/model/Balance.kt
djoksimo: Yes! Thank you for the thoughtful discourse and attention to detail. Nice catch!

sample-app/src/main/java/com/joinforage/android/example/ui/complete/flow/balance/FlowBalanceViewModel.kt
```diff
@@ -71,18 +67,10 @@ class FlowBalanceViewModel @Inject constructor(
...
-
-                    val result = adapter.fromJson(response.data)
-
-                    if (result != null) {
-                        _snap.value = "SNAP: ${result.snap}"
-                        _nonSnap.value = "NON SNAP: ${result.cash}"
+                    val balance = response.toBalance()
+                    if (balance is EbtBalance) {
+                        _snap.value = "SNAP: ${balance.snap}"
+                        _nonSnap.value = "NON SNAP: ${balance.cash}"
```
devinmorgan, line 73: awesome!

devinmorgan (approved):
Nicely done! Thanks for the back and forth -this looks great!

forage-android/src/main/java/com/joinforage/forage/android/network/model/Payment.kt
```diff
@@ -0,0 +1,158 @@
...
+ * @property balance The remaining balance on the EBT Card after the Payment was processed.
+ */
+data class Receipt(
+    val created: String,
+    val ebtCashAmount: String,
+    val isVoided: Boolean,
+    val last4: String,
+    val message: String,
+    val otherAmount: String,
+    val salesTaxApplied: String,
```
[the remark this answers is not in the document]
dleis612: Those fields appear on the object itself, but yes they are unused for standalone Payment objects. The values will always be 0. The choice is:
1. Keep them for consistency sake across other platforms and our docs
2. Remove them because they will always be unused

I lean towards removing them personally.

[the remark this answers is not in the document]
djoksimo: I lean towards removing as well!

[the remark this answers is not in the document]
djoksimo: Done! ea89a177e2daa74b3af03d720a96f31dee978ec0

dleis612 (commented):
The only point I'd like clarity on is the null-safety when parsing each of the models using `jsonObject.get<type>()`. Will this potentially introduce the same issue just seen in iOS in the future?

djoksimo:
> The only point I'd like clarity on is the null-safety when parsing each of the models using `jsonObject.get<type>()`. Will this potentially introduce the same issue just seen in iOS in the future?

@dleis612 Great point! Correct, we'll always have this concern of typed responses + strictness in our mobile SDKs. However, to de-risk this issue:

- I've refactored the internal `GET /payments/` call to only grab the data that is needed (`paymentMethodRef`) for the preliminary requests 
- I've only exposed fields that are meaningful to SDK consumers.
- I've written unit tests in this PR to cover all of the cases where fields are missing to prevent regressions.
- Our E2E tests will cover + prevent issues with "empty/thin" Payments. [Linear ticket](https://linear.app/joinforage/issue/FX-1352/cover-empty-responses-models-in-e2e-tests-for-ios-android)

forage-android/src/main/java/com/joinforage/forage/android/Utils.kt
```diff
@@ -34,3 +35,17 @@ internal fun HttpUrl.Builder.addPathSegmentsSafe(path: String): HttpUrl.Builder
     }
     return this
 }
+
+/**
+ * [JSONObject.optString] has trouble falling back to `null` and seems to fallback to `"null"` (string) instead
+ */
+internal fun JSONObject.getStringOrNull(fieldName: String): String? {
+    if (!has(fieldName)) return null
```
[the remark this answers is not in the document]
djoksimo: 95b630eb4db1cbb104d8a7e18d327bad9a2a2319

dleis612 (approved):
Let's ship it! Looks great

PR #260, opened 2024-06-03
[FX-1368] Re-introduce the ForageVaultWrapper 

evanfreeze:
## What
- Re-introduces the `ForageVaultWrapper`, the `ForagePinSubmitter`, and the new `ForageResponseParser` (to match what we're doing with BT & VGS, even though it doesn't require much translation)

## Why
- Prepping for introducing ForageVault to ecom
- https://linear.app/joinforage/issue/FX-1368/re-introduce-the-foragevaultwrapper-into-the-coreui-module-in-the

## Test Plan
- Tested in sample app that the request flows to rosetta with the correct shape
- Ran unit tests

## How
Can be merged into this feature branch as-is


forage-android/src/main/java/com/joinforage/forage/android/ecom/ui/ForagePINEditText.kt
```diff
@@ -67,16 +67,16 @@ class ForagePINEditText @JvmOverloads constructor(
...
                     // and are thus able to initial LaunchDarkly and find out
-                    // whether to use BT or VGS. So, below we are hedging.
+                    // whether to use BT or Forage. So, below we are hedging.
                     btVaultWrapper = BTVaultWrapper(context, attrs, defStyleAttr)
-                    vgsVaultWrapper = VGSVaultWrapper(context, attrs, defStyleAttr)
+                    forageVaultWrapper = ForageVaultWrapper(context, attrs, defStyleAttr)
                     // ensure all wrappers init with the
                     // same typeface (or the attributes)
-                    btVaultWrapper.typeface = vgsVaultWrapper.typeface
+                    forageVaultWrapper.typeface = btVaultWrapper.typeface
```
djoksimo, line 79:
```suggestion
                    btVaultWrapper.typeface = forageVaultWrapper.typeface
```

should it be this instead

djoksimo (changes requested):

djoksimo (approved):

PR #287, opened 2024-07-11
Move two methods up one level in class hierarchy

devinmorgan:
## What
Move `getVaultSubmitter()` and `showKeyboard()` methods from `ForagePINEditText` to parent `ForagePinElement` class.

## Why
These methods do not care about the specific vault or details that `ForagePINEditText` would specify in Pos or Ecom land. Moving these methods to the parent is a small reducing code duplication / maintenance burden win

## Test Plan
CI tests pass

## Demo
N/A

## How
Can be rolled out immediately

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/core/ui/element/ForagePinElement.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/core/ui/element/ForagePinElement.kt
@@ -7,7 +7,10 @@ import android.view.View
 import android.view.ViewGroup
 import android.widget.LinearLayout
 import com.joinforage.forage.android.R
+import com.joinforage.forage.android.core.services.EnvConfig
 import com.joinforage.forage.android.core.services.VaultType
+import com.joinforage.forage.android.core.services.telemetry.Log
+import com.joinforage.forage.android.core.services.vault.AbstractVaultSubmitter
 import com.joinforage.forage.android.core.ui.VaultWrapper
 import com.joinforage.forage.android.core.ui.element.state.pin.PinEditTextState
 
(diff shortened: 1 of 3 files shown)


devinmorgan:
* **#287** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/287?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @devinmorgan and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

forage-android/src/main/java/com/joinforage/forage/android/core/ui/element/ForagePinElement.kt
```diff
@@ -78,6 +81,13 @@ abstract class ForagePinElement @JvmOverloads constructor(
         vault.clearText()
     }
 
+    override fun showKeyboard() = vault.showKeyboard()
```
evanfreeze, line 84: Can we either adjust `showKeyboard` to be
```kotlin
override fun showKeyboard() {
    vault.showKeyboard()
}
```
or adjust `clearText` to be 
```kotlin
override fun clearText() = vault.clearText()
```

It seems inconsistent to have one be a standard function and the other be assigned with `=`. I don't have a preference for which style we use, just feels like it should be consistent between the two.

devinmorgan: oooooo yeah that's a good point. I'll add that in a follow on-commit before merging. Thanks for flagging

devinmorgan: Just updated to use `{...}`!

evanfreeze (approved):

PR #288, opened 2024-07-11
Make Balance and Capture use sync endpoints

devinmorgan:
## What
Swaps out polling usage in `CapturePaymentRepository` and `CheckBalanceRepository` with sync endpoints by bumping the `API-VERSION` header from `default` -> `2024-01-08`. 

Sources of complexity in this PR:
* Since Rosetta uses the same OkHttpClientBuilder as our model-fetching GET requests use, it was necessary to modify our OkHttp facilities to support specifying a custom `API-VERSION` and using `"default"` as the `API-VERSION` header when left unspecified. Since BT uses its own network facilities, specifying `API-VERSION: 2024-01-08` was less involved.
* BT started returning `gson` representations of JSON instead of ferrying the JSON as a string along so we had to navigate this gracefully

## Why
We've wanted to do this for almost a quarter now in Android. It's already done in iOS and JS

## Test Plan

- ❌ You can manually test the changes in this PR if you want, but I'll be doing a designated bug bash on Friday, July 26th at 10:45 am Eastern.  

## Demo
CI tests pass!

## How
Must be merged after previous PRs in the stack


devinmorgan:
* **#289** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/289?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#288** <a href="https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/288?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @devinmorgan and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BasisTheoryPinSubmitter.kt
```diff
@@ -49,7 +49,7 @@ internal class BasisTheoryPinSubmitter(
         val bt = buildVaultProvider()
 
         val proxyRequest: ProxyRequest = ProxyRequest().apply {
-            headers = vaultProxyRequest.headers
+            headers = vaultProxyRequest.headers + (ForageConstants.Headers.API_VERSION to "2024-01-08")
```
devinmorgan, line 52: Acknowledging that this is ugly. I have plans to clean this up in the future

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
```diff
@@ -15,48 +18,61 @@ internal class BTResponseParser(btRes: Result<Any?>) : VaultResponseParser {
...
+    private fun parseForageError(resRegExp: BtFailureResponseRegExp): ForageApiResponse.Failure? =
+        if (resRegExp.bodyText == null ||
+            resRegExp.statusCode == null ||
+            // if there's a proxy_error, don't try to parse as a Forage error
+            resRegExp.containsProxyError
+        ) {
             null
+        } else {
+            // if there's a body, a status code, and no proxy_error, then
+            // we embrace that the following line will throw an exception
```
devinmorgan, line 53: This exception get's caught in `BtPinSubmitter`, though. It is returned as a `UnknownServerError`

forage-android/src/main/java/com/joinforage/forage/android/core/services/forageapi/network/OkHttpClientBuilder.kt
```diff
@@ -29,43 +30,25 @@ internal object OkHttpClientBuilder {
...
-                            this
+                    .apply {
+                        if (chain.request().headers[ForageConstants.Headers.API_VERSION] == null) {
+                            addHeader(ForageConstants.Headers.API_VERSION, apiVersion)
                         }
-                            // Otherwise, set the default API_VERSION header
-                            ?: addHeader(
-                                ForageConstants.Headers.API_VERSION,
-                                "default"
-                            )
```
devinmorgan, line 41: Opportunity to make the code easier to understand

forage-android/src/main/java/com/joinforage/forage/android/core/services/forageapi/network/OkHttpClientBuilder.kt
```diff
@@ -19,7 +19,8 @@ internal object OkHttpClientBuilder {
         sessionToken: String,
         merchantId: String? = null,
         idempotencyKey: String? = null,
-        traceId: String? = null
+        traceId: String? = null,
+        apiVersion: String = "default"
```
devinmorgan, line 23: `provideOkHttpClient` get's used by a lot of callees. Declaring the `apiVersion` with a param that defaults to `default` will decouple these callers from needing to work directly with `OkHttpClient.addHeader` and the `ForageConstants.Headers`. Less coupling makes for more maintainable code!

forage-android/src/main/java/com/joinforage/forage/android/core/services/forageapi/network/error/payload/ErrorPayload.kt
```diff
@@ -44,7 +44,7 @@ internal abstract class ErrorPayload(
                     return RosettaErrorResponse(jsonErrorResponse)
                 }
                 else -> {
-                    throw UnexpectedResponseError(jsonErrorResponse.toString())
+                    throw UnknownForageFailureResponse(jsonErrorResponse.toString())
```
devinmorgan, line 47: The new name more precisely describes that this exception is thrown when we there is a response that we believe is Forage-defined that we are unable to parse

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
```diff
@@ -15,48 +18,61 @@ internal class BTResponseParser(btRes: Result<Any?>) : VaultResponseParser {
...
-            // via the VaultProxyResponseMonitor. This keeps us informed
-            // of *when* erroroneous BT response happen. Unfortunately,
-            // we do not currently track the specifics of the proxy_error
-            // that that BT returned to us.
-            // TODO: log the specific error that BT responds with when
-            //  resRegExp.containsProxyError == true
-            if (resRegExp.containsProxyError) UnknownErrorApiResponse else null
-        } catch (_: Exception) {
-            null
-        }
```
devinmorgan, line 41: The try catch block is not necessary  here because regexp matching won't throw so I remove it in this PR. To elaborate on why it won't throw:
* if the regexp don't find a match they return null and the compiler enforces that we use null-safety accessors (`?.`).
* if the regexp does match, `.get(0) always refers to the entire match and `.get(1)` refers to the first matched group 

Point being that there won't ever be an index-out-of-bounds scenario by doing `.get(1)` or any other error from not finding a regexp match. I added unit tests in `BTResponseParserText.kt` and `BTFailureResponseRegExp.kt`  to confirm

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
```diff
@@ -15,48 +18,61 @@ internal class BTResponseParser(btRes: Result<Any?>) : VaultResponseParser {
...
-    private fun parseForageError(resRegExp: BaseResponseRegExp): ForageApiResponse.Failure? {
-        if (resRegExp.bodyText == null || resRegExp.statusCode == null) return null
-        return try {
-            ForageApiResponse.Failure(resRegExp.statusCode, resRegExp.bodyText)
-        } catch (_: Exception) {
+    private fun parseForageError(resRegExp: BtFailureResponseRegExp): ForageApiResponse.Failure? =
+        if (resRegExp.bodyText == null ||
+            resRegExp.statusCode == null ||
+            // if there's a proxy_error, don't try to parse as a Forage error
+            resRegExp.containsProxyError
```
devinmorgan, line 48: I replace the try/catch with an additional clause in the if-statement that confirms `resRegExp.containsProxyError == false` before attempting to parse as a ForageError. It is true that attempting to parse an error that is not a Forage error with `ForageApiResponse.Failure(..)` will throw an exception but, if the error is not a proxy_error, then it can only be a ForageError and if we are unable to parse the FroageError, that is an exceptional circumstance and we need to throw.

NOTE: the callee of `BTResponseParser` will catch the exception, log it, and gracefully return `UnknownServerError`

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
```diff
@@ -15,48 +18,61 @@ internal class BTResponseParser(btRes: Result<Any?>) : VaultResponseParser {
...
+            // the response as the value of Any? resolves to a
+            // com.google.gson.internal.LinkedTreeMap instance.
+            // Fortunately, this is of type Map<Any, Any?> so we can
+            // still parse it with JSONObject and avoid having to
+            // directory depend on Gson ourselves
+            val result = vaultResponse.getOrThrow()
+            val jsonResponse = when {
+                (result == null) -> JSONObject()
+                (result == "") -> JSONObject()
+                (result is Map<*, *>) -> JSONObject(result)
```
devinmorgan, line 71: For whatever reason, using API-VERSION: 2024-01-08 caused BT to start returning a Gson representation of the JSON response. While this is a bad oversight of BT to leak a third-party dependency in their public API, fortunately, the Gson representation can be cast as a `Map<*, *>` and the built-int `JSONObject` constructor correctly parses.

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
```diff
@@ -15,48 +18,61 @@ internal class BTResponseParser(btRes: Result<Any?>) : VaultResponseParser {
...
+            // Basis Theory appears to leak their Gson dependency in
+            // the response as the value of Any? resolves to a
+            // com.google.gson.internal.LinkedTreeMap instance.
+            // Fortunately, this is of type Map<Any, Any?> so we can
+            // still parse it with JSONObject and avoid having to
+            // directory depend on Gson ourselves
+            val result = vaultResponse.getOrThrow()
+            val jsonResponse = when {
+                (result == null) -> JSONObject()
+                (result == "") -> JSONObject()
```
devinmorgan, line 70: These are to gracefully handle the deferred pin-submission flows since they technically have an empty body and were throwing with the new `API-VERSION: 2024-01-08`

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BasisTheoryPinSubmitter.kt
```diff
@@ -61,7 +63,18 @@ internal class BasisTheoryPinSubmitter(
...
+        } catch (e: UnknownBTSuccessResponse) {
+            logger.e("Unknown success response from BasisTheory Vault.", e)
+            UnknownErrorApiResponse
+        } catch (e: UnknownForageFailureResponse) {
+            logger.e("Unknown error from Payments API.", e)
+            UnknownErrorApiResponse
+        } catch (e: Exception) {
+            logger.e("Request to Basis Theory failed.", e)
+            UnknownErrorApiResponse
+        }
```
devinmorgan, line 77: and this is where we catch the various exceptions that could arise from parsing BT responses

forage-android/src/test/java/com/joinforage/forage/android/network/data/CapturePaymentRepositoryTest.kt
devinmorgan, line 1: remove tests that tested whether polling was explicitly used or not, as polling no longer happens with sync endpoints. This is a separate endeavor from dropping all the polling code, which is done in the next PR

forage-android/src/test/java/com/joinforage/forage/android/vault/BasisTheoryPinSubmitterTest.kt
```diff
@@ -52,7 +52,7 @@ class BasisTheoryPinSubmitterTest() : MockServerSuite() {
         mockBasisTheoryTextElement = mock(TextElement::class.java)
 
         // ensure we don't make any live requests!
-        mockBasisTheoryResponse(Result.success("success"))
+        mockBasisTheoryResponse(Result.success(mapOf("mySuccessfulKey" to "mySuccessfulValue")))
```
devinmorgan, line 55: Updating because BT starts returning Gson representation of JSON instead of just the JSON string

forage-android/src/main/java/com/joinforage/forage/android/core/services/vault/CapturePaymentRepository.kt
```diff
@@ -38,37 +37,24 @@ internal class CapturePaymentRepository(
...
-            operationDescription = "payment capture of Payment $paymentRef"
         )
-        if (pollingResponse is ForageApiResponse.Failure) {
-            return pollingResponse
-        }
 
-        return when (val paymentResponse = paymentService.getPayment(paymentRef)) {
+        return when (response) {
             is ForageApiResponse.Success -> {
                 logger.i("[HTTP] Received updated Payment $paymentRef for Payment $paymentRef")
```
dleis612, line 54: Can we update the log statement now that we don't do a GET for the `Payment`? Also lol that we include the payment ref twice in this string

forage-android/src/main/java/com/joinforage/forage/android/core/services/vault/CheckBalanceRepository.kt
```diff
@@ -40,31 +40,28 @@ internal class CheckBalanceRepository(
...
-        val vaultResponse = when (
+        val balanceResponse = when (
             val response = vaultSubmitter.submit(
                 getVaultRequestParams(encryptionKeys, paymentMethod)
             )
         ) {
-            is ForageApiResponse.Success -> Message.ModelMapper.from(response.data)
+            // response comes as (snap, non_snap) but we've historically
+            // returned (snap, cash) in our SDK public API. So, we need
+            // to parse the JSON nad convert to (snap, cash) here
```
dleis612, line 50: nit: 'and'

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
dleis612: A learning I took from implementing these changes in iOS and JS is that we should really force all error scenarios through the BT codepath to ensure we are happy with the parsing. The error scenarios are:
- EBT network error code (eg. `ebt_error_55`)
- Forage API error (eg. `invalid_merchant_header`)
- BT error (eg. IDK COULD BE ANYTHING)

dleis612 (changes requested):

forage-android/src/main/java/com/joinforage/forage/android/core/services/vault/CapturePaymentRepository.kt
devinmorgan: great callout about updating the log statement! Ha! I didn't see that we double-captured `paymentRef` until just now :P

forage-android/src/main/java/com/joinforage/forage/android/core/services/vault/CheckBalanceRepository.kt
devinmorgan: rut ro I don't think I follow 😭. can you elaborate on the callout?

forage-android/src/main/java/com/joinforage/forage/android/ecom/services/vault/bt/BTResponseParser.kt
devinmorgan: Great point! I'll incorporate this into the bug bash on Friday. Thanks for surfacing this!!

forage-android/src/main/java/com/joinforage/forage/android/core/services/vault/CheckBalanceRepository.kt
devinmorgan: type = fixed! going to mark as resolved!

forage-android/src/main/java/com/joinforage/forage/android/core/services/vault/CapturePaymentRepository.kt
devinmorgan: Your feedback showed that we can actually drop `when` blocks in both the `CapturePaymentRepository` class and the `CheckBalanceRepository` class. Now that I've done that, I'll mark this thread as resolved

dleis612 (approved):

devinmorgan:
### Merge activity

* **Jul 24, 4:41 PM EDT**: @devinmorgan merged this pull request with [Graphite](https://app.graphite.dev/github/pr/teamforage/forage-android-sdk/288).

PR #297, opened 2024-07-29
[TW-700] Remove mid/ from Android online-only SDK docs 

kimberleehowley:
This PR updates the `merchantId` description and code snippets to reflect the new best practices after PAID-1227 (to no longer use `mid/`).


devinmorgan (approved):
I'm really excited about this PR, and it definitely has my +1. My understanding is that we're holding off on merging this in until the latest version of the SDKs will play nicely with it. If so, I leave it to you to merge this in when appropriate!

PR #313, opened 2025-07-10
[EXPA-207] Small fixes and refactorings

marcparmet:
## What
Small fixes and refactorings.

In preparation for larger changes for upcoming credit card payment sheet.

## Why
HSA/FSA for GoPuff

## Test Plan
No new unit test coverage needed.

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ecom/services/ForageSDK.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ecom/services/ForageSDK.kt
@@ -135,7 +135,7 @@ class ForageSDK {
         return tokenizeService.tokenizeCard(
             cardNumber = foragePanEditText.getPanNumber(),
             customerId = customerId,
-            reusable = reusable ?: true
+            reusable = reusable
         )
     }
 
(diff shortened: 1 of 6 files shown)


forage-android/src/main/java/com/joinforage/forage/android/ecom/services/ForageSDK.kt
```diff
@@ -135,7 +135,7 @@ class ForageSDK {
         return tokenizeService.tokenizeCard(
             cardNumber = foragePanEditText.getPanNumber(),
             customerId = customerId,
-            reusable = reusable ?: true
+            reusable = reusable
```
devinmorgan, line 138: My understanding is that `resusable ?: true` will default to `true` if the client does not specify and that `resuable` alone is `None` which defaults to `false`. I'm just a little worried that our existing partners may rely on not passing anything defaults to re-usable cards and that this change would break that. Am I missing something?

marcparmet: `TokenizeEBTCardParams.reusable` is the source here and it's declared as `Boolean`, not `Boolean?`, so the Elvis operation is not necessary. My IDE even grays it out.
Screenshot 2025-07-10 at 9 46 31 AM

devinmorgan: I have peace of mind, now! Thanks for your patience with me, Marc!

PR #316, opened 2025-07-10
[EXPA-208] Remove ForageConfigManager

marcparmet:
## What
Remove ForageConfigManager.

This functionality is now obsolete. With only Rosetta in use and no Datadog initialization in either ForagePanElement or ForagePinElement, ForageConfigManager adds unnecessary indirection and no longer provides value.

## Why
HSA/FSA for GoPuff

## Test Plan
Existing tests provide coverage.


forage-android/src/main/java/com/joinforage/forage/android/core/ui/element/ForagePanElement.kt
```diff
@@ -188,11 +188,6 @@ abstract class ForagePanElement @JvmOverloads constructor(
             }
     }
 
-    private val forageConfigManager = ForageConfigManager {
-            forageConfig ->
-        initWithForageConfig(forageConfig)
```
devinmorgan, line 193: Unless I'm missing something, we still need to call `initWithForageConfig` (line 219/220 of this file) on the first invocation of `setForageConfig`. I was think this PR would basically rehome `ForageConfigManager`'s logic  into this file (e.g. the `ForagePanElement` class), to eliminate the separate `ForageConfigManager` class without removing the logic—essentially flattening the structure.

Am I on the same page as you?

devinmorgan: all I had to do was scroll down ... sorry for the confusion 🤦

devinmorgan (approved):
This is great!

PR #326, opened 2025-08-07
[EXPA-241] Introduce ForagePaymentSheetField

marcparmet:
## What
Introduce ForagePaymentSheetField.

This design better reflects what we have in our iOS SDK.

## Why
HSA/FSA for GoPuff

## Test Plan
Unit testing

### DIFF
--- a/forage-android/src/main/java/com/joinforage/forage/android/ecom/ui/element/ForagePaymentSheet.kt
+++ b/forage-android/src/main/java/com/joinforage/forage/android/ecom/ui/element/ForagePaymentSheet.kt
@@ -10,8 +10,8 @@ import android.util.AttributeSet
 import android.view.inputmethod.InputMethodManager
 import android.widget.LinearLayout
 import androidx.annotation.ColorInt
+import androidx.annotation.VisibleForTesting
 import androidx.constraintlayout.widget.ConstraintLayout
-import androidx.core.content.ContextCompat.getSystemService
 import androidx.core.content.getSystemService
 import com.google.android.material.textfield.TextInputEditText
 import com.google.android.material.textfield.TextInputLayout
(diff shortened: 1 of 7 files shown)


forage-android/src/main/java/com/joinforage/forage/android/ecom/ui/element/ForagePaymentSheet.kt
```diff
@@ -56,76 +67,59 @@ class ForagePaymentSheet @JvmOverloads constructor(
...
-        securityCodeEditText = findViewById(R.id.securityCodeEditText)
-        zipCodeEditText = findViewById(R.id.zipCodeEditText)
-
-        allEditText = setOf(
-            cardholderNameEditText,
-            cardNumberEditText,
-            expirationEditText,
-            securityCodeEditText,
-            zipCodeEditText
+        val cardholderNameEditText: TextInputEditText = findViewById(R.id.cardholderNameEditText)
```
amuletxheart, line 82: (talking) `findViewById` is a pretty old school way of doing things, we should probably use one of the newer patterns instead. For this PR as long as we're consistent with the existing codebase it should be fine.

forage-android/src/main/java/com/joinforage/forage/android/ecom/ui/element/ForagePaymentSheet.kt
```diff
@@ -134,115 +128,270 @@ class ForagePaymentSheet @JvmOverloads constructor(
...
+) : ForageElement<ElementState>, EditTextElement {
+    protected var _onChangeEventListener: StatefulElementListener<ElementState>? = null
+
+    private var onFocusEventListener: SimpleElementListener? = null
+    private var onBlurEventListener: SimpleElementListener? = null
+    private var focusState = FocusState.forEmptyInput()
+
+    // There's no getStroke() method on GradientDrawable so we have to cache
+    // initial values from editable_frame.xml.
+    internal var boxStrokeColor = 0xFFD0D0D0.toInt()
```
amuletxheart, line 189: (talking) can we move the color constant to an xml file?

